PR #1, opened 2018-07-29
add feature-gated serde support

apoelstra:


dongcarl:
Most likely needs rebase, I'm guessing #2 blocks this?

sgeisler:
Maybe I'm missing something but the serde unit tests don't seem to run on travis. Maybe there should be a second tests run with activated serde `cargo test --features serde`?

apoelstra:
Oops, you're right! I added that but in a branch that I later abandoned.

src/serde_macros.rs
```diff
@@ -0,0 +1,42 @@
...
+            fn deserialize<D: ::serde::Deserializer<'de>>(d: D) -> Result<$t, D::Error> {
+                use ::serde::de::Error;
+                use hex::FromHex;
+
+                if d.is_human_readable() {
+                    let sl: &str = ::serde::Deserialize::deserialize(d)?;
+                    $t::from_hex(sl).map_err(D::Error::custom)
+                } else {
+                    let sl: &[u8] = ::serde::Deserialize::deserialize(d)?;
+                    if sl.len() != $t::len() {
```
dongcarl: Can we use `from_slice` here?

src/sha256d.rs
```diff
@@ -112,6 +113,24 @@ input: &'static str,
             assert_eq!(hash, manual_hash);
         }
     }
+
+    #[cfg(feature="serde")]
+    #[test]
+    fn sha256_serde() {
+        use serde_test::{Configure, Token, assert_tokens, assert_ser_tokens, assert_de_tokens};
+
+        static HASH_BYTES: [u8; 32] = [
```
dongcarl: Why is static needed here?

src/serde_macros.rs
```diff
@@ -0,0 +1,42 @@
+
+#[cfg(feature="serde")]
+macro_rules! serde_impl(
```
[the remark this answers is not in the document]
sgeisler: Since we noticed that we might be able to use `1.24.0` we can probably refactor out the `$len` parameter.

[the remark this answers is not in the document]
apoelstra: On further reflection I think we probably don't want to go too far past `1.19` which is the currently-supported mrustc version. Though `1.20` is not too bad, it only requires one additional layer of bootstrapping to avoid "trusting trust" attacks; this would give us associated constants at least.

src/sha256d.rs
apoelstra: just an optimization

PR #3, opened 2018-07-29
Add hash160 and sha512 (needed for bip32)

apoelstra:


src/hash160.rs
```diff
@@ -25,7 +25,7 @@ use {Error, Hash};
 
 /// Output of the Bitcoin HASH160 hash function
 #[derive(Copy, Clone, PartialEq, Eq)]
-pub struct Hash160Hash([u8; 20]);
+pub struct Hash160Hash(pub [u8; 20]);
```
ariard: What is the reason of getting the internal array public ? Doesn't seems clear to me at first sight ?

apoelstra: It makes conversions to/from `[u8;20]` more natural. I don't think allowing this is very footgunny, and without it, it's very annoying de/serializing hashes.

PR #9, opened 2018-11-08
hex: allow specifying a width when printing hashes

apoelstra:


src/hex.rs
```diff
@@ -84,26 +84,43 @@ impl<'a> Iterator for HexIterator<'a> {
...
-        write!(fmt, "{:02x}", *ch)?;
+pub fn format_hex(data: &[u8], f: &mut fmt::Formatter) -> fmt::Result {
+    let width = f.width().unwrap_or(2 * data.len());
+    for _ in (2 * data.len())..width {
+        f.write_str("0")?;
+    }
+    for ch in data.into_iter().take(width / 2) {
+        write!(f, "{:02x}", *ch)?;
+    }
+    if width < 2 * data.len() && width % 2 == 1 {
```
sgeisler: Is it actually useful to support nibble-wise writing instead of byte-wise (I find it a little bit hard to reason about, but no big problem if you say it's useful)?

apoelstra: The reason is only that bytewise would be confusing for a user, who would otherwise request `n` bytes and get `2n` characters.

I agree it's not really useful, the goal is just to be intuitive.

sgeisler: Yeah, that seems like a worthwhile goal. Then I'll review this version.

src/hex.rs
```diff
@@ -84,26 +84,43 @@ impl<'a> Iterator for HexIterator<'a> {
...
 /// efficient than going through a `String` using `ToHex`.
-pub fn format_hex<T: fmt::Write>(data: &[u8], mut fmt: T) -> fmt::Result {
-    for ch in data {
-        write!(fmt, "{:02x}", *ch)?;
+pub fn format_hex(data: &[u8], f: &mut fmt::Formatter) -> fmt::Result {
+    let width = f.width().unwrap_or(2 * data.len());
+    for _ in (2 * data.len())..width {
+        f.write_str("0")?;
+    }
+    for ch in data.into_iter().take(width / 2) {
```
sgeisler: This is cutting off the least significant bytes if width is too short :/ That's not what you'd expect imo. E.g. integers aren't truncated if width is too short:

```rust
assert_eq!(&format!("{:02}", 100u8), "100");
assert_eq!(&format!("{:05}", 100u8), "00100");

assert_eq!(&format!("{:02x}", 4u8), "04");
assert_eq!(&format!("{:01x}", 4u8), "4");
assert_eq!(&format!("{:01x}", 42u8), "2a");
```

I think we should do the same as the standard library here.

apoelstra: Which bytes in a hash are less significant than others?

If we can't truncate there's no point to this PR.

sgeisler: You are right, that terminology makes not much sense, but still, it's rather arbitrary that the last bytes get cut off instead of the first ones. And it would also be unexpected/non-standard behavior as all the other `LowerHex` impls don't truncate.

This PR still adds the padding functionality, which is a good thing imo (although I'd think about allowing user defined fill characters and alternate formatting with `0x` prepended for it to be as [standard](https://doc.rust-lang.org/std/fmt/index.html) as possible).

dongcarl: I believe what we want is precision rather than width.

## Excerpts from the Rust docs ([link](https://doc.rust-lang.org/std/fmt/#width))
### Width
This is a parameter for the "_minimum_ width" that the format should take up.
### Precision
For ___non-numeric___ types, this can be considered a "_maximum_ width".

apoelstra: Ah, OK, thanks.

apoelstra:
Rebased and added commit to separate out "width" from "precision".

@sgeisler can you re-review?

PR #11, opened 2018-12-04
rename all the hash types to `Hash`

apoelstra:


src/lib.rs
```diff
@@ -54,7 +54,7 @@ pub use error::Error;
 /// A hashing engine which bytes can be serialized into. It is expected
 /// to implement the `io::Write` trait, but to never return errors under
 /// any conditions.
-pub trait HashEngine: Clone + io::Write {
```
sgeisler: I personally like `HashEngine` better since it at least tells you *something* about it's function.

src/ripemd160.rs
```diff
@@ -23,28 +23,30 @@ use std::io;
...
-use {Error, Hash, HashEngine};
+use Engine as EngineTrait;
+use Hash as HashTrait;
+use Error;
 
 const BLOCK_SIZE: usize = 64;
 
 /// Engine to compute RIPEMD160 hash function
-pub struct Ripemd160Engine {
+pub struct Engine {
```
sgeisler: I'd rather rename it to `HashEngine`. I'm sure someone will just import `some_hash::*` and then I'll be annoyed to not really know what an `Engine` is. It's much better to know "this is an engine for the only hash function used in this module" than "this is an engine (steam, electrical, hash, idk)".

apoelstra:
The real goal is to reduce naming redundancy. I hate typing

```
use bitcoin_hashes::sha2::Sha2Hash;
use bitcoin_hashes::sha512::Sha512Hash;
```
etc., when I could just type
```
use bitcoin_hashes::{sha2, sha512};
// use sha2::Hash, sha512::Hash
```
and save a pile of typing while also making my types visually more distinct from other identifiers by their `lowercase::CamelCase` pattern, and also making it easier for people using multiple hashlibs who only want to import `bitcoin_hashes`.

sgeisler:
lgtm

PR #13, opened 2018-12-13
Implement borrowing as `&[u8]` for Hashes à la Vec

dongcarl:


src/lib.rs
```diff
@@ -77,7 +77,7 @@ pub trait Hash: Copy + Clone + PartialEq + Eq +
     ops::Index<ops::RangeTo<usize>, Output = [u8]> +
```
sgeisler: Aren't the `Index` impls kinda redundant if you have `Borrow<[u8]>`?

dongcarl: Don't think so... We can have both, just like `Vec` does.

sgeisler: Ok, makes sense if you think of it as a slim wrapper around a byte array.

PR #14, opened 2018-12-13
Implement into_inner for all Hash types

sgeisler:
Even though the inner value is public for all hash types except `Hmac` it's better to have a well defined common interface for getting the underlying data structure.

I'd propose to make the inner values private to enforce usage of `into_inner(self)`.

## TODO:
- [x] Make inner structs private
- [ ] Tests

### DIFF
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -36,6 +36,7 @@ serde_impl!(Hash, 20);
 
 impl HashTrait for Hash {
     type Engine = sha256::HashEngine;
+    type Inner = [u8; 20];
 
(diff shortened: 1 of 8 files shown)


dongcarl:
Perhaps I shouldn't have merged this as making `Hash.0` not `pub` is a breaking change. @apoelstra thoughts?

apoelstra:
Yeah, I guess we'll have to make our next release 0.2.0. No biggie, I don't think we had any non-breaking changes in the pipeline that we wanted to get through first (and actually I think we don't have any users yet anyway).

PR #25, opened 2019-01-17
Use write_all instead of write

dongcarl:


sgeisler:
What is the reason for this PR? #21 made some changes to the implicit API contract. Does this have something to do with it? Is the library currently broken and we didn't catch it? If so we'd need more test cases for that.

Generally the changes look good. For an ACK I still have to check if there aren't places that also need to be changed to `write_all` and were forgotten.

dongcarl:
Since the explicit API contract for `write` never promises to write all the data, and our goal in these lines is to write all the data, I believe that regardless of the implicit contract we should be using `write_all`.

sgeisler:
@dongcarl I definitely agree. My question is: was anything broken by #21 because it relied on the implicit contract and we didn't notice?

dongcarl:
@sgeisler I don't think so, unless you've noticed something I haven't.

PR #29, opened 2019-01-23
Deserialize into owned String for human-readable deserializer

dongcarl:
- Modify tests to properly roundtrip for serde (de)serialization.

This blocks https://github.com/rust-bitcoin/rust-bitcoin/pull/215


dongcarl:
We need a new release after this, as this is blocking integration into rust-bitcoin and will cause panics for packages dependent on `bitcoin_hashes`.

PR #30, opened 2019-01-23
Remove extraneous block_size method from Hash trait

dongcarl:
Since `HashEngine` already has a `block_size` method, and each `Hash` has its corresponding engine under `T::Engine`, we don't need a `block_size` method on `Hash`

### DIFF
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -56,10 +56,6 @@ impl HashTrait for Hash {
         20
     }
 
-    fn block_size() -> usize {
-        64
-    }
-
     fn from_slice(sl: &[u8]) -> Result<Hash, Error> {
         if sl.len() != 20 {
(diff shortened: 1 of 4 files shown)


sgeisler:
Is this related to #22 (haven't looked into it more in more depth yet)? If yes I'd like to avoid just going back and forth all the time.

Since we can use 1.22 now, shouldn't we replace these methods with associated constants to get rid of some of the [`debug_assert`s](https://github.com/rust-bitcoin/bitcoin_hashes/pull/30/files#diff-c76d1ccc0023b9b98f395f6c6da93b16R43)?

dongcarl:
@sgeisler See: https://github.com/rust-bitcoin/bitcoin_hashes/pull/31

dongcarl:
Closing in favor of just merging #31

PR #35, opened 2019-02-06
Allow initializing a SHA256 engine with a state

### LINKED ISSUE #34: Allow initializing engine with midstate
Right now it seems like callers can not set the state of hash engines, f.e. SHA256. Sometimes it's useful to initialize a hash engine "as if" you've already hashed some data. If you've obtained the midstate this can speed up hashing significantly in some cases.

jonasnick:
Fixes #34 (for SHA256)

### DIFF
--- a/src/sha256.rs
+++ b/src/sha256.rs
@@ -72,6 +72,21 @@ index_impl!(Hash);
 serde_impl!(Hash, 32);
 borrow_slice_impl!(Hash);
 
+impl Hash {
+    /// Creates an engine with the given `midstate` and `length` of already
+    /// hashed input data. This function should not be used directly unless you
+    /// really know what you're doing.
+    pub fn engine_with_state(midstate: [u8; 32], length: usize) -> HashEngine {
+        let mut h = [0; 8];
+        BigEndian::read_u32_into(&midstate, &mut h);
+        HashEngine {
+            h: h,
+            length: length,
+            buffer: [0; BLOCK_SIZE],
+        }
+    }
+}
+
 impl HashTrait for Hash {
     type Engine = HashEngine;
     type Inner = [u8; 32];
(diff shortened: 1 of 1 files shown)


sgeisler:
utACK, but I kinda dislike that it's not part of the `Hash` or `HashEngine` trait. That makes the API less predictable: why does sha256 have this function but no other hash implementation? They are all able to extract the midstate.

jonasnick:
> why does sha256 have this function but no other hash implementation?

The main application I'm interested in is taproot hashing which can use tagged hashing a la `H(H("taproot") || H("taproot") || rest)` and presumably uses SHA256. This PR can certainly be pushed back until there's concrete proposal with community support.

sgeisler:
After experimenting with different generalizations a bit I came to the conclusion that we can't make `from_midstate` part of the `HashEngine` trait without some refactorings/conceptual changes. 

The main problem is that I can't find a good way to implement `from_midstate`for HMAC since it would need a key to initialize the outer engine. Maybe the midstate of the HMAC engine should be 
```rust
(<<T as Hash>::Engine as HashEngine>::MidState, <<T as Hash>::Engine as HashEngine>::MidState)
```
so it could be actually used to reconstruct the hash engine?

apoelstra:
@sgeisler +1 to that

stevenroose:
I agree that `HashEngine::from_midstate` would be more intuitive for this than a method on `Hash`. It also helps hiding the method from the average user, fwiw. Also using `HashEngine::Midstate` as input instead of `[u32; 8]` would be preferred, even though that probably makes it less efficient to create engines with it.

stevenroose:
For now I have this PR cherry-picked in https://github.com/rust-bitcoin/bitcoin_hashes/pull/39.

If we come up with a way to include `from_midstate` in the trait, that'd be great. But for now I can test the taproot hashes and midstate usage for elements using #39.

apoelstra:
Merged as part of #39

PR #36, opened 2019-03-02
fix non-human-readable serde deserialization

apoelstra:


apoelstra:
IMHO this requires only a minor version bump since it strictly increases the set of things that can be deserialized.

Prior to this PR we would attempt to deserialize hashes first as `&[u8]`, which we would copy into the hash type. The [Deserialize impl for &[u8]](https://github.com/serde-rs/serde/blob/master/serde/src/de/impls.rs#L486-L493) uses a `Visitor` object that can only handle "borrowed bytes" (which are slices with a sufficiently long lifetime that they can be returned to callers; compare "bytes", which are also borrowed, but which are not assumed to outlive the invocation of `Visitor::visit_bytes`).

We don't need our byteslices to live long enough to to be returned, since we just copy the bytes into a new `Hash` object and return that. So we implement our own visitor which does exactly the same thing as before, but implements `visit_bytes` rather than `visit_borrowed_bytes` (and the default impl of `visit_borrowed_bytes` will fallthrough to `visit_bytes`).

This is needed to roundtrip using serde_cbor and probably other binary deserializiers.

src/serde_macros.rs
```diff
@@ -17,21 +17,63 @@ macro_rules! serde_impl(
...
+                    struct HexVisitor;
+
+                    impl<'de> ::serde::de::Visitor<'de> for HexVisitor {
+                        type Value = $t;
+
+                        fn expecting(&self, formatter: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+                            formatter.write_str("an ASCII hex string")
+                        }
+
+                        fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
```
dongcarl: I don't know `serde` well enough, but unless there's something that _requires_ `visit_bytes` we can just implement `visit_str` here?

apoelstra: The `String` deserializer implements `visit_bytes`, so I didn't want to break anything depending on that.

PR #39, opened 2019-04-28
Add sha256::Midstate type

stevenroose:
Builds on https://github.com/rust-bitcoin/bitcoin_hashes/pull/40/.

In Elements we use this type to represent witness merkle roots and Asset IDs.

### DIFF
--- a/src/sha256.rs
+++ b/src/sha256.rs
@@ -209,6 +209,20 @@ macro_rules! round(
 );
 
 impl HashEngine {
+    /// Create a new [HashEngine] from a midstate.
+    ///
+    /// Be aware that this method only works when [length] is a multiple of the block size.
+    pub fn from_midstate(midstate: Midstate, length: usize) -> HashEngine {
+        let mut ret = [0; 8];
+        BigEndian::read_u32_into(&midstate[..], &mut ret);
+
+        HashEngine {
+            buffer: [0; BLOCK_SIZE],
+            h: ret,
+            length: length,
+        }
+    }
+
     // Algorithm copied from libsecp256k1
     fn process_block(&mut self) {
         debug_assert_eq!(self.buffer.len(), BLOCK_SIZE);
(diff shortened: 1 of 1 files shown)


src/sha256.rs
```diff
@@ -129,6 +130,59 @@ impl HashTrait for Hash {
...
+}
+
+impl hex::FromHex for Midstate {
+    /// Parses a hex string as a hash object
+    fn from_hex(s: &str) -> Result<Self, Error> {
+        if s.len() != 2 * Self::LEN {
+            return Err(Error::InvalidLength(2 * Self::LEN, s.len()));
+        }
+
+        Ok(Self(hex::FromHex::from_hex(s)?))
```
apoelstra: this is a compile error on 1.22

src/sha256.rs
```diff
@@ -129,6 +130,59 @@ impl HashTrait for Hash {
...
+borrow_slice_impl!(Midstate);
+
+impl Midstate {
+    /// Length of the midstate, in bytes.
+    const LEN: usize = 32;
+
+    /// Flag indicating whether user-visible serializations of this hash
+    /// should be backward. For some reason Satoshi decided this should be
+    /// true for `Sha256dHash`, so here we are.
+    const DISPLAY_BACKWARD: bool = false;
```
apoelstra: I don't think this constant makes sense outside of a trait.

src/sha256.rs
```diff
@@ -129,6 +130,59 @@ impl HashTrait for Hash {
...
+
+    /// Unwraps the [Midatate] and returns the underlying byte array.
+    pub fn into_inner(self) -> [u8; 32] {
+        self.0
+    }
+
+    /// Produce a [Midstate] from the current state of a SHA246 engine.
+    pub fn from_sha256_engine(e: HashEngine) -> Midstate {
+        Midstate(e.midstate())
+    }
```
apoelstra: I'd prefer this function not exist, since we don't want users to think of midstates as first-class objects.

src/sha256.rs
```diff
@@ -129,6 +130,59 @@ impl HashTrait for Hash {
...
+        if sl.len() != Self::LEN {
+            Err(Error::InvalidLength(Self::LEN, sl.len()))
+        } else {
+            let mut ret = [0; 32];
+            ret.copy_from_slice(sl);
+            Ok(Midstate(ret))
+        }
+    }
+
+    /// Unwraps the [Midatate] and returns the underlying byte array.
```
apoelstra: typeo Midatate

apoelstra:
You need to change the type of `sha256::Engine::Midstate` to the new type.

src/sha256.rs
stevenroose: the fmt_hex macro uses it.. It's private.

src/sha256.rs
stevenroose: I can compile on 1.22 locally after I updated.

src/sha256.rs
stevenroose: Removed and had `sha256::Hash::midstate` return the type directly.

src/sha256.rs
stevenroose: fixed

src/sha256.rs
```diff
@@ -133,6 +134,62 @@ impl HashTrait for Hash {
     }
 }
 
+/// Output of the SHA256 hash function
+#[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
+pub struct Midstate([u8; 32]);
```
apoelstra: I think we should make the internal representation public. It's a bit more footgunny but would make the taproot midstates much simpler to implement.

src/sha256.rs
```diff
@@ -133,6 +134,62 @@ impl HashTrait for Hash {
...
+        if s.len() != 2 * Self::LEN {
+            return Err(Error::InvalidLength(2 * Self::LEN, s.len()));
+        }
+
+        // Self::DISPLAY_BACKWARD is true
+        let mut vec = Vec::<u8>::from_hex(s)?;
+        vec.reverse();
+        Self::from_slice(&vec)
+    }
+}
```
apoelstra: We can now replace this with the more efficient implementation.

src/sha256.rs
stevenroose: How is that different from `from_inner`?

src/sha256.rs
stevenroose: pushed

src/sha256.rs
apoelstra: `from_inner` is a trait method and is available in generic code

stevenroose: Should we then not do that for all Hash types? In fact that would be helpful in some situations where we are using "the one hash" up to "the n hash" for special purposes. (F.e. asset ID and reissuance token calculation.)

stevenroose: > `from_inner` is a trait method and is available in generic code

That's also not true. The `from_inner` method on `Midstate` is not a trait method.

stevenroose: To conclude, I think we can publish the internal representation until `const fn` is a stable thing, then we can break and have people use `from_inner` instead.

src/sha256.rs
```diff
@@ -209,6 +209,20 @@ macro_rules! round(
 );
 
 impl HashEngine {
+    /// Create a new [HashEngine] from a midstate.
+    ///
+    /// Be aware that this method only works when [length] is a multiple of the block size.
```
apoelstra: Can you `assert!` this? Ok if you add it in a separate commit @stevenroose

apoelstra:
ACK except nit

src/sha256.rs
stevenroose: Done.

stevenroose:
Added blocksize multiple assertion in separate commit.

PR #40, opened 2019-04-28
Add Hash::from_inner and prevent double allocation in FromHex using it

stevenroose:


src/hex.rs
```diff
@@ -84,6 +97,40 @@ impl<'a> Iterator for HexIterator<'a> {
...
+        let len = self.sl.len() / 2;
+        (len, Some(len))
+    }
+}
+
+impl<'a> DoubleEndedIterator for HexIterator<'a> {
+    fn next_back(&mut self) -> Option<Result<u8, Error>> {
+        let current_len = self.sl.len();
+        if current_len % 2 == 1 {
+            Some(Err(Error::OddLengthString(self.sl.len())))
```
stevenroose: Why not use `current_len` here?

src/hex.rs
```diff
@@ -84,6 +97,40 @@ impl<'a> Iterator for HexIterator<'a> {
...
+        } else if self.sl.is_empty() {
```
stevenroose: Why not use `current_len` here as well? I would start with `if current_len == 0 { ... } else if current_len % 2 == 1 { ... }` personally.

stevenroose: But I see this is mirroring the impl of `Iterator`, which if fine I guess.

stevenroose:
Can't approve my own PR. So ACK d71dd9959bd72626f245fdc6ed01d5b0a21db4db

src/hex.rs
```diff
@@ -82,6 +97,40 @@ impl<'a> Iterator for HexIterator<'a> {
...
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let len = self.sl.len() / 2;
+        (len, Some(len))
+    }
+}
+
+impl<'a> DoubleEndedIterator for HexIterator<'a> {
+    fn next_back(&mut self) -> Option<Result<u8, Error>> {
+        let current_len = self.sl.len();
+        if current_len % 2 == 1 {
```
sgeisler: Imo this check shouldn't happen for every element, but I'll open a separate PR.

src/hex.rs
sgeisler: I took the liberty to fix it in a separate commit, do you think I should squash it into Andrew's commit?

PR #42, opened 2019-05-02
Tagged SHA256 hashes

stevenroose:
Introducing a new `SHA256t` hash type that works as a tagged hash. The idea of tagged hashes is to prevent collisions between hashes meant for different purposes. This type allows tagging these hashes using Rust's type system so that they are semantically different from each other and can't be confused.

It also enables seamless efficient calculation of these hashes.

The `Tag::engine()` method should return a `HashEngine` that is prefilled with data. Of course, it is supposed to have this data statically initialized instead of calculated on call.

Example usage: https://github.com/rust-bitcoin/rust-bitcoin/pull/259

### DIFF
--- a/src/sha256t.rs
+++ b/src/sha256t.rs
@@ -152,3 +152,34 @@ impl<'de, T: Tag> ::serde::Deserialize<'de> for Hash<T> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use ::{Hash, sha256, sha256t};
+    use ::hex::ToHex;
+
+    #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Default, Hash)]
+    pub struct TestHashTag;
+
+    impl sha256t::Tag for TestHashTag {
+        fn engine() -> sha256::HashEngine {
+            // The TapRoot TapLeaf midstate.
+            let midstate = sha256::Midstate::from_inner([
+               156, 224, 228, 230, 124, 17, 108, 57, 56, 179, 202, 242, 195, 15, 80, 137, 211, 243,
+               147, 108, 71, 99, 110, 96, 125, 179, 62, 234, 221, 198, 240, 201,
+            ]);
+            sha256::HashEngine::from_midstate(midstate, 64)
+        }
+    }
+
+    /// A hash tagged with `$name`.
+    pub type TestHash = sha256t::Hash<TestHashTag>;
+
+    #[test]
+    fn test_sha256t() {
+       assert_eq!(
+           TestHash::hash(&[0]).to_hex(),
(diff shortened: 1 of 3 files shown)


dr-orlovsky:
Is there anything that prevents this from being merged?

apoelstra:
I'd prefer there be a real system out there (even a branch of elements where we can launch a regtest node) so we can test the API in the wild.

dr-orlovsky:
We are using tagged hashes in RGB: https://github.com/LNP-BP/lnpbps/blob/master/lnpbp-0001.md
and it will be nice to call it from here: https://github.com/LNP-BP/rust-lnpbp/blob/9a9d09e52b5f50eaa3b84cdca91cabd584cf77c8/src/cmt/pubkey.rs#L67

src/sha256t.rs
```diff
@@ -0,0 +1,148 @@
...
+#[cfg(feature="serde")]
+impl<'de, T: Tag> ::serde::Deserialize<'de> for Hash<T> {
+    fn deserialize<D: ::serde::Deserializer<'de>>(d: D) -> Result<Hash<T>, D::Error> {
+        use ::hex::FromHex;
+
+        if d.is_human_readable() {
+            struct HexVisitor;
+
+            impl<'de> ::serde::de::Visitor<'de> for HexVisitor {
+                type Value = Hash<T>;
```
dr-orlovsky: Build with `std-serde` feature fails here: <https://travis-ci.org/rust-bitcoin/bitcoin_hashes/jobs/640164066#L323>

It also happens locally on my machine with nightly version + `std-serde`:

```
error[E0401]: can't use generic parameters from outer function
  --> src/sha256t.rs:98:35
   |
90 | impl<'de, T: Tag> ::serde::Deserialize<'de> for Hash<T> {
   |           - type parameter from outer function
91 |     fn deserialize<D: ::serde::Deserializer<'de>>(d: D) -> Result<Hash<T>, D::Error> {
   |        ----------- try adding a local generic parameter in this method instead
...
98 |                 type Value = Hash<T>;
   |                                   ^ use of generic parameter from outer function
```

I tried different ways of fixing the problem: introducing another generic parameter right on `HexVisitor` impl etc - nothing works here

dr-orlovsky: Have fixed it finally: <https://github.com/stevenroose/bitcoin_hashes/pull/3>

stevenroose:
Thanks, @dr-orlovsky! This PR should be good for review then!

apoelstra:
Can you add a unit test which actually implements this `Tag` trait so it's clear how to use it?

stevenroose:
I added the tapleaf hash as an example.

PR #43, opened 2019-05-02
hmac: make `Midstate` type be something we can actually construct an engine from

apoelstra:


src/hmac.rs
```diff
@@ -29,6 +29,25 @@ use {Error, Hash, HashEngine};
 #[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
 pub struct Hmac<T: Hash>(T);
 
+/// Pair of underlying hash midstates which represent the current state
+/// of an `HmacEngine`
+pub struct HmacMidState<T: Hash> {
+    inner: <T::Engine as HashEngine>::MidState,
```
stevenroose: Why make them private? If so, I would at least have `into_` methods. Either `into_inner` and `into_outer` or either `into_inner_outer() -> (inner, outer)` or something. But I think it's easier to just have the fields public.

apoelstra: Sure, I'll just make them public

stevenroose: Also public midstate fields makes it easier to construct static midstate objects.

PR #46, opened 2019-05-14
Add siphash24 module

stevenroose:
Based on the siphasher crate (https://github.com/jedisct1/rust-siphash/) that is a copy of the implementation in `std::hash`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,6 +45,7 @@ pub mod sha1;
 pub mod sha256;
 pub mod sha512;
 pub mod sha256d;
+pub mod siphash24;
 pub mod cmp;
 
 use std::{borrow, fmt, hash, io, ops};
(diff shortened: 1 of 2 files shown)


apoelstra:
Is there a significant performance increase from using unsafe code rather than just letting `byteorder` do the endianness conversions?

apoelstra:
Also, is there another implementation we can test against? Or some test vectors?

stevenroose:
To be honest, I have no idea about the implementation internals. I just took the one from the siphasher package. I also took their unit tests, but I had to adapt them since they were using the `std::hash` trait or something.

About more unit tests, I will look for some in other languages that can be verified line per line because I understand the ones I have aren't verbatim from siphasher and that might sound phishy :)

stevenroose:
@apoelstra hehe, I noticed that the test vectors we have are actually the same ones as the reference implementation has: https://github.com/veorq/SipHash/blob/master/vectors.h#L3

I think the siphasher people made a variation on that (so they could use the Hash trait) and I coincidentally reverted their variation back to the original vectors. It's a series of 64 hashes of the series `[]`, `[0]`, `[0, 1]`, `[0, 1, 2]`, ...

PR #47, opened 2019-05-20
hex: Simplify and DRY-ify HexIterator

dongcarl:
Old logic can be _much_ simpler if we make the `HexIterator` wrap a `Char` iterator instead.

### DIFF
--- a/src/hex.rs
+++ b/src/hex.rs
@@ -16,6 +16,7 @@
 //!
 
 use std::fmt;
+use std::str::Chars;
 use {Error, Hash};
 
 /// Trait for objects that can be serialized as hex strings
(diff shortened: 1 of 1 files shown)


src/hex.rs
```diff
@@ -63,78 +64,59 @@ impl<T: Hash> FromHex for T {
...
+    /// The `Chars` iterator whose first two characters will be decoded to yield
+    /// the next byte.
+    chars: Chars<'a>,
 }
 
 impl<'a> HexIterator<'a> {
-    /// Constructs a new `HexIterator` from a string slice. If the string is of odd length it
-    /// returns an error.
+    /// Constructs a new `HexIterator` from a string slice. If the string is of
+    /// odd length it returns an error.
```
stevenroose: Nice! Someone else who likes comments to be within 80 columns! Do you do this manually or do you have a Vim plugin that allows different textwidths for code vs comments?

dongcarl: Haha I use emacs and Meta-q seems to just auto-align the current comment block you're in :grin:

Issue #45, opened 2019-05-12
hashes: Missing utility functions


dpc:
`as_ref()`, `as_mut()`, `to_vec()`, `as_slice()`

stevenroose:
Do you mean for accessing the inner types?

dpc:
Yes.

stevenroose:
I think they are supposed to be immutable. F.e. they have `Copy`.

You have the `into_inner` method that probably is the best way to get the inner type and together with `from_inner`, it can also be mutated.

stevenroose:
Also, they implement the various indexing traits, so you can do `&my_hash[..]` to get the underlying slice.

dpc:
I don't understand what are you trying to say. For stuff like some generics some people might need `AsRef` one way or another. The other ones would be just convenient to have so importing `AsRef` is not neccessary. `into_inner` is nice, but what if I have `&GivenHash` for some reason?

I guess the mutability part is about not wanting the Hashes to be accidentally mutated? Fair enough.

apoelstra:
+1 to `Deref<[u8]>` and `AsRef<[u8]>`.

I see no reason to create `to_vec` though I think we might get that for free from the above; and I don't want any mutators.

stevenroose:
AsRef is already implemented. I implemented Deref in https://github.com/rust-bitcoin/bitcoin_hashes/pull/49.

apoelstra:
Closed by #49

PR #49, opened 2019-06-13
Implement std::ops::Deref<Target=Inner> for Hash impls

stevenroose:
Relevant for #45.

### DIFF
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -33,7 +33,7 @@ hex_fmt_impl!(Display, Hash);
 hex_fmt_impl!(LowerHex, Hash);
 index_impl!(Hash);
 serde_impl!(Hash, 20);
-borrow_slice_impl!(Hash);
+borrow_slice_impl!(Hash, [u8; 20]);
 
 impl HashTrait for Hash {
     type Engine = sha256::HashEngine;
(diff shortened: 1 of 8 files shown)


apoelstra:
Can you add a unit test that confirms that this gives us `to_vec`? Are there other things, e.g. the indexing traits, that we can now drop?

dpc:
As per https://doc.rust-lang.org/std/ops/trait.Deref.html

"Implementing Deref for smart pointers makes accessing the data behind them convenient, which is why they implement Deref. On the other hand, the rules regarding Deref and DerefMut were designed specifically to accommodate smart pointers. **Because of this, Deref should only be implemented for smart pointers to avoid confusion.**"

I'm not saying that I agree/disagree, but just want to make sure you're aware of the official party line.

apoelstra:
Oh, yeah, I think when I approved this I had it in my head that we were implementing `Deref<[u8]>` (i.e. "hashes are kind of like a slice"), which I think is consistent with the party line. For example `Vec<T>` implements `Deref<[T]>`. Implementing `Deref<[u8]>` should get us all the functionality requested in #45.

Agreed that `Deref<[u8; N]>` is not want we want.

apoelstra:
I'm going to release a new major version of this lib today/next week to get `no_std` support. We can change the `Deref` impl then.

PR #50, opened 2019-06-23
increase version to 0.4.0

apoelstra:


apoelstra:
cc @dongcarl @stevenroose

stevenroose:
Nice! With these changes we could do the taproot tagged hashes as a minor change.

PR #51, opened 2019-06-28
fix hex iterator

apoelstra:


src/sha256.rs
```diff
@@ -66,6 +67,13 @@ impl EngineTrait for HashEngine {
 #[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
 pub struct Hash([u8; 32]);
 
+impl str::FromStr for Hash {
```
stevenroose: Can't FromStr be implemented for all `Hash` trait types? Or even for all `FromHex` types?

src/hex.rs
```diff
@@ -64,9 +64,9 @@ impl<T: Hash> FromHex for T {
 
 /// Iterator over a hex-encoded string slice which decodes hex and yields bytes.
 pub struct HexIterator<'a> {
-    /// The `Chars` iterator whose first two characters will be decoded to yield
+    /// The `Bytes` iterator whose next two bytes will be decoded to yield
     /// the next byte.
-    chars: Chars<'a>,
+    iter: str::Bytes<'a>,
```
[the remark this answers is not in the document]
stevenroose: I think the motivation was something with multi-byte characters that mess up the `.len()` check for oddness in the beginning.

src/hex.rs
```diff
@@ -97,21 +97,21 @@ impl<'a> Iterator for HexIterator<'a> {
...
-        let lo = self.chars.next().unwrap();
+        let hi = self.iter.next()?;
+        let lo = self.iter.next().unwrap();
         Some(chars_to_hex(hi, lo))
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let len = self.chars.as_str().len() / 2;
-        (len, Some(len))
+        let (min, max) = self.iter.size_hint();
```
stevenroose: Could this be the culprit of the performance hit?

It uses `slice::Iter`'s size_hint, which uses the following macro, not sure what to make of it:
```rust

// To get rid of some bounds checks (see `position`), we compute the length in a somewhat
// unexpected way. (Tested by `codegen/slice-position-bounds-check`.)
macro_rules! len {
    ($self: ident) => {{
        let start = $self.ptr;
        let diff = ($self.end as usize).wrapping_sub(start as usize);
        let size = mem::size_of(start);
        if size == 0 {
            diff
        } else {
            // Using division instead of `offset_from` helps LLVM remove bounds checks
            diff / size
        }
    }}
}
```

src/hex.rs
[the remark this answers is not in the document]
sgeisler: Ah, right, we were using `.len()` insted of `.chars().count()` (it's kinda unintuitive that they aren't the same).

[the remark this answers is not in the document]
apoelstra: It's because `.len()` is O(1) while `chars().count()` is O(n).

I am pretty surprised that the byte iterator is slower than the chars iterator, especially by such a large factor. I suspect this is a bug in stdlib.

src/sha256.rs
apoelstra: No, because `T: Hash` includes types outside our crate, and `str::FromStr` is not our trait :(.

src/hex.rs
apoelstra: Hmm, I don't think so.

PR #52, opened 2019-07-05
Add `no_std` support

### LINKED ISSUE #12: no-std mode
Imo the core functionality of this library could be implemented without the std lib which would enable it to be used in embedded contexts (HW wallets?). For example `byteorder` does this by having a (default) feature `std`, which can be deactivated if needed.

Am I missing something here? If not I think it's a worthwhile goal.

apoelstra:
When `no_std` is on we lose the `io::Write` impls on all the hash engines, naturally, as well as the `std::error::Error` impl on `Error`.

We also lose the `ToHex` trait, since this outputs a `String`, though we retain the `format_hex` function as well as `fmt::{Display, LowerHex, UpperHex}` for all hashes.

Fixes #12


src/std_impls.rs
```diff
@@ -0,0 +1,161 @@
...
+use {sha1, sha256, sha512, ripemd160, siphash24};
+use HashEngine;
+use Error;
+
+impl error::Error for Error {
+    fn cause(&self) -> Option<&error::Error> { None }
+    fn description(&self) -> &str { "`std::error::description` is deprecated" }
+}
+
+impl io::Write for sha1::HashEngine {
```
sgeisler: That's a lot of identical code. Would a macro be appropriate or are we trying to get rid of them?

src/siphash24.rs
```diff
@@ -135,14 +134,23 @@ impl HashEngine {
...
 
-impl Write for HashEngine {
-    #[inline]
-    fn flush(&mut self) -> ::std::io::Result<()> {
-        Ok(())
+impl EngineTrait for HashEngine {
+    type MidState = State;
+
+    #[cfg(not(feature = "fuzztarget"))]
+    fn midstate(&self) -> State {
```
sgeisler: Although this seems to be just a move I'd be interested if the conditional compilation makes any sense here (I don't see the point of it).

.gitignore
```diff
@@ -7,3 +7,5 @@ Cargo.lock
 #fuzz
 fuzz/hfuzz_target
 fuzz/hfuzz_workspace
+
+*~
```
sgeisler: I should have complained earlier when the `.idea` stuff got added, but these editor-specific ignores would be better suited for `.git/info/exclude` since we don't have a blessed standard editor and don't want to add new ones all the time.

src/lib.rs
```diff
@@ -27,15 +27,18 @@
 #![deny(unused_mut)]
 #![deny(missing_docs)]
 
+#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
 #![cfg_attr(all(test, feature = "unstable"), feature(test))]
 #[cfg(all(test, feature = "unstable"))] extern crate test;
 
+#[cfg(any(test, feature="std"))] extern crate core;
```
sgeisler: Don't we want to include `core` just all the time?

src/lib.rs
```diff
@@ -27,15 +27,18 @@
 #![deny(unused_mut)]
 #![deny(missing_docs)]
 
+#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
```
sgeisler: We should really add an "integration test" that is not a test and doesn't use `std` to see if it actually compiles. I'm not sure what's the best way to do this. Maybe we don't need to treat tests in a special way since `std` is active by default (so tests will never run without `std` anyway in practice)?

.gitignore
apoelstra: I don't mind adding new ones all the time :P. Do you think we'll ever have a use for files ending in `~`?

In any case I also don't mind removing this.

src/siphash24.rs
apoelstra: Heh, yeah, I'll get rid of it. It does look weird.

src/std_impls.rs
apoelstra: Heh, I couldn't decide. It's only 40 or 50 lines.

src/lib.rs
apoelstra: Oh, yeah, we should add `--no-default-features` to travis.

Compiling with `cargo test` compiles twice, once with `#[cfg(test)]` and once without. (That's why errors are doubled up usually.)

src/lib.rs
apoelstra: No, when `no_std` is on then it's implicit.

.gitignore
sgeisler: Well, then it's ok I guess ^^

apoelstra:
Rebased, fixed build failure with `--features "fuzztarget"`, added no_std build test to Travis

apoelstra:
Removed redundant fuzztarget/nonfuzztarget `siphash24::midstate`

stevenroose:
Is there a specific reason why you implemented input for each hash instead of using a macro like there was for `write_impl`? Could have prevented a bit of code duplication.

Otherwise ACK.

apoelstra:
It was just short, and all the duplication was in one file right next to each other. If I need to change it I'll convert it to a macro.

apoelstra:
Added a commit to bump the version number.

@sgeisler am I ok to merge?

elichai:
If anyone really wants the `ToHex` trait, they could try and maybe make it similar to this https://github.com/debris/rustc-hex/blob/master/src/lib.rs#L31
(generic over `FromIterator<char>`)

apoelstra:
cherry-picked both. Thanks!

dongcarl:
ACK eac6cde4e0ad9957fc4a2745b4668a9880339d8c

PR #56, opened 2019-07-10
sha1: Correct fuzzing behaviour.

dongcarl:
This makes the behavior of sha1's midstate function in line with everything else, which it should be since it also `write_impl!`s

We should look into how to make our files for individual hashes more DRY using macros or something else, as inconsistencies like this seem to crop up a lot when people forget to modify one file or another and is annoying.


apoelstra:
You cannot use macros to replace individual trait method implementations, unfortunately.

PR #57, opened 2019-07-11
Repository cleanup w/ help from Clippy

dongcarl:
Various cleanups inspired by clippy output. Added sane clippy allows in the last commit. `cargo clippy` should have empty output after this is merged.

### DIFF
--- a/src/hex.rs
+++ b/src/hex.rs
@@ -127,7 +127,7 @@ pub fn format_hex(data: &[u8], f: &mut fmt::Formatter) -> fmt::Result {
     for _ in (2 * data.len())..width {
         f.write_str("0")?;
     }
-    for ch in data.into_iter().take(prec / 2) {
+    for ch in data.iter().take(prec / 2) {
         write!(f, "{:02x}", *ch)?;
     }
     if prec < 2 * data.len() && prec % 2 == 1 {
--- a/src/hmac.rs
+++ b/src/hmac.rs
@@ -54,6 +54,12 @@ pub struct HmacEngine<T: HashTrait> {
     oengine: T::Engine,
 }
 
+impl<T: HashTrait> Default for HmacEngine<T> {
+    fn default() -> Self {
+        HmacEngine::new(&[])
+    }
+}
+
 impl<T: HashTrait> HmacEngine<T> {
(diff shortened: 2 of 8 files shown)


stevenroose:
Hmm, your clippy annotations broke the 1.22 build. Perhaps for now we better just drop them? Or comment them or something for reference.

Otherwise ACK fee48f0b46958f257df2963ee51a4341254b397b

apoelstra:
Are we supposed to "not take references if it's small enough to pass by value" or "use `iter` rather than `into_iter` to avoid moving a 2-word value"? :)

apoelstra:
Less joking aside, I think

* We should drop the `into_iter` commit because to the extent it has any effect at all, it's just introducing an unnecessary indirection. If Clippy recommends this it's a bug in clippy.
* We should drop the "take `siphash` by value" commit because it's a minor thing and the only breaking change in this PR. (Though I suppose making `engine` a trait method of `Hash` is technically a breaking change as well, if somebody had implemented `Hash` on their own type and implemented an `engine` method on that which will now conflict.)
* We should drop the clippy-lint commit because it breaks compilation. Though I'm glad to learn that in recent versions of rustc these are no longer compile errors; that prevented me using Clippy for a long time.

dongcarl:
@apoelstra

> We should drop the `into_iter` commit because to the extent it has any effect at all, it's just introducing an unnecessary indirection. If Clippy recommends this it's a bug in clippy.

I believe this lint is for readability, as we expect `into_` methods to move, but it clearly doesn't in this case. But happy to drop if you still think it should be dropped. Source: https://rust-lang.github.io/rust-clippy/master/index.html#into_iter_on_ref

> We should drop the "take `siphash` by value" commit because it's a minor thing and the only breaking change in this PR.

Sure, let's do that.

> We should drop the clippy-lint commit because it breaks compilation. Though I'm glad to learn that in recent versions of rustc these are no longer compile errors; that prevented me using Clippy for a long time.

Done.

> We may need to drop `derive(Clone)` on sha512 because in the distant past (not sure about 1.22) `Clone` was not implemented for `[u8; 128]`.

It is implemented for `1.22.0`, I don't think we need to drop this.

apoelstra:
Ok, "for readability" makes sense. All looks good to me then.

PR #58, opened 2019-07-15
Create a separate hex::Error type

stevenroose:


apoelstra:
utACK. Can you add a commit which bumps the major version number and adds a changelog entry?

stevenroose:
Done.

PR #59, opened 2019-08-13
Add hash_newtype macro

stevenroose:
Adds a publicly exported macro that allows for easily creating newtypes around bitcoin_hashes hash types.

This helps with https://github.com/rust-bitcoin/rust-bitcoin/issues/284.


apoelstra:
Can you add a minor version bump (and CHANGELOG.md entry) to this PR?

src/serde_macros.rs
```diff
@@ -64,11 +64,9 @@ macro_rules! serde_impl(
                             E: ::serde::de::Error,
                         {
                             if v.len() != $t::LEN {
-                                Err(E::invalid_length(v.len(), &stringify!($len)))
+                                Err(E::invalid_length(v.len(), &format!("{}", $t::LEN).as_str()))
```
apoelstra: AFAIK `stringify!` produces a string literal while this code will allocate a string

src/serde_macros.rs
```diff
@@ -64,11 +64,9 @@ macro_rules! serde_impl(
...
                             } else {
-                                let mut ret = [0; $len];
-                                ret.copy_from_slice(v);
-                                Ok($t(ret))
+                                Ok($t::from_slice(v).expect("length already checked"))
```
apoelstra: At least nominally this adds an extra error path

src/serde_macros.rs
stevenroose: Correct. Not sure how rustc can optimize that allocation away as it knows the `LEN` is constant. Anyway I thought since it's in an error path, performance was not so important. I could change this back with the effect that users of `hash_newtype` need to mention the hash length in the macro. Which I'm also fine with, btw. Let me know if you prefer this reverted.

src/serde_macros.rs
stevenroose: Yeah this is not related to removing the `$len` matcher, but because the newtypes have a `Hash` as inner type and not a `[u8; $len]`.

One thing I thought about is to `map_err` the `from_slice` error into the serde error which would avoid a panic in the scenario where `from_slice` adds another error possibility.

stevenroose:
> Can you add a minor version bump (and CHANGELOG.md entry) to this PR?

Done.

src/serde_macros.rs
apoelstra: I'd prefer it reverted

apoelstra: Actually, I'm not so concerned about allocations in error paths ... it's the other change that I want reverted, but I see that that's difficult.

stevenroose: Reverted.

src/serde_macros.rs
stevenroose: I changed it to use map_err, to at least exclude a panic possibility.

PR #61, opened 2019-09-05
sha256: from_midstate can take reference

stevenroose:


apoelstra:
I generally prefer to take short bytestrings by value to avoid indirections. Especially as these types are `Copy` in all cases (I think) anyway.

elichai:
Yep. with `Copy` types passing by value can even be faster than by reference.

stevenroose:
I'm actually curious about how this works. What are the exact differences in the effects of using references or Copy for those types. What are "indirections" and what (performance?) problem do they pose?

I'm relatively new to lower-level programming like this.

elichai:
> I'm actually curious about how this works. What are the exact differences in the effects of using references or Copy for those types. What are "indirections" and what (performance?) problem do they pose?
> 
> I'm relatively new to lower-level programming like this.

Here's an example https://godbolt.org/z/295sHh you can see that the first one starts with a `mov` instruction and according to `llvm-mca` is one cycle less.
you can also see that the second one actually does no memory reads, `lea` edits the memory in-place without reading it, while the other one needs to move from the reference to a local and then add to it itself.

Disclaimer: my x86 is a bit rusty, so I might be a little off here in explaining `lea`

stevenroose:
Yeah but a u32 is a single word. `[u8; 32]` is several words. Anyhow, I'll close it. godbolt is interesting, but it's hard to get snippets concise enough that are still interesting to investigate. All simple things I tried with `[u8; 32]` were just introspection so it was optimized to just be inlined it seemed and just treat it as a reference always.

elichai:
> Yeah but a u32 is a single word. `[u8; 32]` is several words. Anyhow, I'll close it. godbolt is interesting, but it's hard to get snippets concise enough that are still interesting to investigate. All simple things I tried with `[u8; 32]` were just introspection so it was optimized to just be inlined it seemed and just treat it as a reference always.

Right. it was just an example on how it can even happen. more complex types can trigger more complex llvm passes that decide to choose between move and reference.

PR #63, opened 2019-09-25
Fixing `hash_newtype` macro to work correctly

dr-orlovsky:
While I was working on implementing https://github.com/rust-bitcoin/rust-bitcoin/issues/284 I've found that the new `hash_newtype` can't be used outside of `bitcoin_hashes` project due to fact that macros used by it are not exported. This PR fixes the problem for most of such macros.

~~Nevertheless, I am still not able to run https://github.com/pandoracore/rust-bitcoin/blob/hashtypes/src/lib.rs#L67 due to problems with `serde` crate access from inside `hash_newtype` macro. Any suggestions?~~

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,6 +157,18 @@ macro_rules! hash_newtype {
             }
         }
 
+        impl ::std::convert::From<$hash> for $newtype {
+            fn from(inner: $hash) -> Self {
+                Self::from_inner($hash)
+            }
+        }
+
+        impl ::std::convert::To<$hash> for $newtype {
+            fn to(&self) -> $hash {
+                Self::to_inner()
+            }
+        }
+
         impl $newtype {
             /// The length in bytes.
             pub const LEN: usize = <$hash as $crate::Hash>::LEN;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,9 +31,9 @@
 #![cfg_attr(all(test, feature = "unstable"), feature(test))]
 #[cfg(all(test, feature = "unstable"))] extern crate test;
 
-#[cfg(any(test, feature="std"))] extern crate core;
-#[cfg(feature="serde")] extern crate serde;
-#[cfg(all(test,feature="serde"))] extern crate serde_test;
+#[cfg(any(test, feature="std"))] pub extern crate core;
+#[cfg(feature="serde")] pub extern crate serde;
+#[cfg(all(test, feature="serde"))] extern crate serde_test;
 extern crate byteorder;
 
 #[macro_use] mod util;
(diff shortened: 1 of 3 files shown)


thomaseizinger:
You should be able to solve that by re-exporting the whole serde crate from `bitcoin_hashes` through `pub extern crate serde` and using it inside the macro as `$crate::serde`.

dr-orlovsky:
Thank you, @thomaseizinger, that has worked

dr-orlovsky:
Some builds are failed due to https://github.com/rust-bitcoin/bitcoin_hashes/pull/64

This PR fixing the problem has to be merged first before getting this PR merged

stevenroose:
One possible downside IMO is that it seems also our internal macros are exported. Does anyone know if it's possible to just export certain macros?

dr-orlovsky:
Basically only macros with `#[macro_export]` are exported, so it's not that we export internal macros. However, I had to do all previously internal macros used by `hash_newtype` macro to be exportable, otherwise `hash_newtype` can't be exportable itself.

So not all internal macros are exported, but the list of the exported macros has grown

src/lib.rs
```diff
@@ -31,9 +31,9 @@
 #![cfg_attr(all(test, feature = "unstable"), feature(test))]
 #[cfg(all(test, feature = "unstable"))] extern crate test;
 
-#[cfg(any(test, feature="std"))] extern crate core;
-#[cfg(feature="serde")] extern crate serde;
-#[cfg(all(test,feature="serde"))] extern crate serde_test;
+#[cfg(any(test, feature="std"))] pub extern crate core;
```
stevenroose: Does it make sense to export core? It's not a third party dep, right?

dr-orlovsky:
Ok, with the new macOS build process is broken again even with all those changes, so I need to figure out how to fix it. The actual problem happens with C compilation of fsevents library, used by browserify and babel, and AFAIR the dependencies I made was a solution to fix it.

So let's keep this PR closed and when I will solve npm install bugs for new mac OS I will just open a new PR

src/lib.rs
dr-orlovsky: You are right, it can be removed. Done

dr-orlovsky: No, I have to bring it back, otherwise the hash_newtype results in 
```
error[E0603]: crate `core` is private
 --> <::hashes::util::hex_fmt_impl macros>:2:18
```

stevenroose:
I'm not sure if we want the newtypes to implement `Hash`.. Opinions?

dr-orlovsky:
I did it since otherwise a lot of code in rust-bitcoin has to be modified in order to get Txid working correctly. What are the arguments of not implementing Hash trait over the hash types?

apoelstra:
I think the newtypes should implement `Hash`.

PR #66, opened 2019-11-27
Fix travis

elichai:
Travis lately is failing with `cargo web`,
it seems to be related to a bump in llvm in beta which apparently was fixed here: https://github.com/rust-lang/rust/commit/2bf59bea481dd4b4365cafe2e94fa8bf330a6877

IIUC the reason it ran on beta is so it won't clog up the stable run (because it already has fuzzing there)
https://github.com/rust-bitcoin/bitcoin_hashes/compare/aa468e7de40379fd6e9a909ec8fc6b004ca0aac1..785b57b1f8e7a6b90089e682cc85850b1f1f99f7

So I splitted the jobs here to have the regular stable+beta+nightly+MSRV jobs, and on top of those stable+wasm and stable+fuzzing.
I also enabled caching like I did here: https://github.com/rust-bitcoin/rust-secp256k1/pull/148 so that the install of `cargo-web` won't take 10 minutes.

@TheBlueMatt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,39 @@
 language: rust
+cache: 
+  cargo: true
+  directories:
+  - cargo_web
+
 rust:
   - stable
   - beta
   - nightly
   - 1.22.0
 
+jobs:
+  include:
+    - rust: stable
+      env: FUZZ=true
+    - rust: stable
+      env: WASM=true
+
+
 before_install:
   - sudo apt-get -qq update
   - sudo apt-get install -y binutils-dev libunwind8-dev
 
 script:
-  - cargo build --verbose
-  - cargo test --verbose
-  - cargo test --verbose --features "serde"
-  - cargo build --verbose --features "fuzztarget"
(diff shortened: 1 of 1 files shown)


elichai:
On my own travis, 
First run: "19 min 47 sec" (no caching yet)  https://travis-ci.org/elichai/bitcoin_hashes/builds/617818528
Second run: "9 min 23 sec" (with caching) https://travis-ci.org/elichai/bitcoin_hashes/builds/617827509

apoelstra:
Turns out this broke Travis. The build has been broken on 1.29 for a while and Travis has been failing to tell us.

elichai:
> Turns out this broke Travis. The build has been broken on 1.29 for a while and Travis has been failing to tell us.

:(
Because of the missing `&&` between the commands, fuck.

apoelstra:
Lol, well, my lesson is to not somehow go an entire year without manually testing any PRs on 1.29 in this library.

PR #67, opened 2019-11-29
Bump version to 0.7.2

apoelstra:


apoelstra:
cc @dr-orlovsky I dropped off IRC, but my understanding is that none of these changes are breaking. In particular, although the `hash_newtype` macro is now completely different, there were no possible users before since you couldn't make a usage of the macro compile :P.

cc @TheBlueMatt can I have an ack?

dr-orlovsky:
Yep, I do agree

dr-orlovsky:
Sorry, just have found that my rebase from master has cleared an important part of the code: https://github.com/rust-bitcoin/bitcoin_hashes/pull/68

Can you pls accept that PR before doing the version bump and publishing to crates.io?

apoelstra:
Ok, rebased

PR #68, opened 2019-11-30
Adding back crate exports required by hash_newtype macro

dr-orlovsky:
They were missed due to the last auto merge from the master


dr-orlovsky:
Also there was a rustc 1.22-compiler specific "bug" (a new syntax used which is not supported by 1.22) and this leads to failing the AS_DEPENDENCY builds for rust-bitcoin. Now it's fixed in the commit in this PR

PR #69, opened 2019-12-13
Add `as_hash` method to the hash_newtype macro

stevenroose:


stevenroose:
I changed the AsRef to an explicit `as_hash` impl. It's less useful generically, though.. But it can't hurt, I guess. It makes it possible to convert between two newtypes in a single line without explicit generics.

PR #71, opened 2020-01-08
Make hex path in serde macros relative to the crate

elichai:
Currently the serde macros (use in creating new hash types) did:
`use hex::ToHex;`
which works fine in this library but not anywhere else.
the only reason it works in `rust-bitcoin` is because it actually used the `hex` crate(https://docs.rs/hex/0.4.0/hex/trait.ToHex.html) without anyone noticing (which was left in non-test code by accident) 

so now it should use the correct hex traits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitcoin_hashes"
-version = "0.7.3"
+version = "0.7.4"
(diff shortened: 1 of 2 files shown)


elichai:
The only test that fails is the Emscripten one, I'll try to fix it but it's unrelated to this PR.

PR #72, opened 2020-01-10
Add from_hash to the newtype API

stevenroose:
Should have been added together with `as_hash`. Just for convenience, I don't need a release for this.


dr-orlovsky:
I'm up for it!

PR #73, opened 2020-01-22
Cleaning up macro namespaces

dr-orlovsky:
I tried to apply the same rules to all macros:
* for identities from outside of the crate, belonging to `std` and `core` use `::std::...` and `::core::..`
* for all other, use `$crate::`

Rationale: I had run an issue when compiler says `no core` while importing some macro and using it in the mods that do not import core directly.

Related / additional info: https://github.com/rust-bitcoin/rust-bitcoin/pull/395

### DIFF
--- a/src/hex.rs
+++ b/src/hex.rs
@@ -201,9 +201,9 @@ macro_rules! impl_fromhex_array {
     ($len:expr) => {
         impl $crate::hex::FromHex for [u8; $len] {
             fn from_byte_iter<I>(iter: I) -> Result<Self, $crate::hex::Error>
-                where I: ::std::iter::Iterator<Item=Result<u8, $crate::hex::Error>> +
-                    ::std::iter::ExactSizeIterator +
-                    ::std::iter::DoubleEndedIterator,
+                where I: ::core::iter::Iterator<Item=Result<u8, $crate::hex::Error>> +
+                    ::core::iter::ExactSizeIterator +
+                    ::core::iter::DoubleEndedIterator,
             {
                 if iter.len() == $len {
                     let mut ret = [0; $len];
--- a/src/hex.rs
+++ b/src/hex.rs
@@ -199,11 +199,11 @@ impl FromHex for Vec<u8> {
 
 macro_rules! impl_fromhex_array {
     ($len:expr) => {
-        impl FromHex for [u8; $len] {
-            fn from_byte_iter<I>(iter: I) -> Result<Self, Error>
-                where I: Iterator<Item=Result<u8, Error>> +
-                    ExactSizeIterator +
-                    DoubleEndedIterator,
+        impl $crate::hex::FromHex for [u8; $len] {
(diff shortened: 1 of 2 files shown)


stevenroose:
Hmm, can you refer to some argumentation elsewhere that argues this practice? I have seen `$crate::std::` and `$crate::core::` before as well.

stevenroose:
Ok, I think I agree on the `::core` and `::std` convention. 

It seems like we're using `::std::iter` instead of `::core::iter` somewhere, though.

dr-orlovsky:
@stevenroose Fixed. BTW what is the difference between `core` and `std` iterators? Docs seems to be exactly the same for both...

dr-orlovsky:
Ok, I was wrong. When you use `::core` in macro inside `bitcoin_hashes`, and than it is called from macro in `rust-bitcoin`, it will fail:

```
error[E0433]: failed to resolve: unresolved import
  --> <::bitcoin_hashes::util::borrow_slice_impl macros>:3:36
   |
1  |    ($ ty : ident) => (borrow_slice_impl ! ($ ty,) ;) ;
   |    -                  ----------------------------- in this macro invocation (#3)
   |   _|
   |  |
2  |  | ($ ty : ident, $ ($ gen : ident : $ gent : ident), *) =>
3  |  | (impl < $ ($ gen : $ gent), * > :: core :: borrow :: Borrow < [u8] > for $ ty
   |  |                                    ^^^^
   |  |                                    |
   |  |                                    unresolved import
   |  |                                    help: a similar path exists: `hashes::core`
4  |  |  < $ ($ gen), * > { fn borrow (& self) -> & [u8] { & self [..] } } impl < $
...   |
9  |  |      type Target = [u8] ; fn deref (& self) -> & Self :: Target { & self . 0 }
10 |  |  })
   |  |   -
   |  |   |
   |  |___in this expansion of `borrow_slice_impl!` (#2)
   |      in this expansion of `borrow_slice_impl!` (#3)
   | 
  ::: src/hash_types.rs:59:1
   |
59 |    hash_newtype!(FilterHash, sha256d::Hash, 32, doc="Bloom filter double-SHA256 locator hash, as defined in BIP-168");
   |    ------------------------------------------------------------------------------------------------------------------- in this macro invocation (#1)
   | 
  ::: <::bitcoin_hashes::hash_newtype macros>:1:1
   |
1  | /  ($ newtype : ident, $ hash : ty, $ len : expr, $ docs : meta) =>
2  | |  {
3  | |      # [$ docs] #
4  | |      [derive (Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)] pub
...  |
8  | |      borrow_slice_impl ! ($ newtype) ; impl $ newtype
   | |      --------------------------------- in this macro invocation (#2)
...  |
33 | |      }
34 | |  } ;
   | |____- in this expansion of `hash_newtype!` (#1)
```

And hundreds of errors like that.

PR #76, opened 2020-01-27
Implement FromStr for hash newtypes

stevenroose:
Seems to have been an oversight. @apoelstra Let me know if you want me to bump the version right away.

@dr-orlovsky can you review?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -192,6 +192,13 @@ macro_rules! hash_newtype {
                 self.0.into_inner()
             }
         }
+
+        impl ::std::str::FromStr for $newtype {
+            type Err = <$hash as ::std::str::FromStr>::Err;
+            fn from_str(s: &str) -> ::std::result::Result<$newtype, Self::Err> {
+                s.parse().map($newtype)
+            }
+        }
     };
 }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -192,6 +192,13 @@ macro_rules! hash_newtype {
                 self.0.into_inner()
             }
         }
+
+        impl ::std::str::FromStr for $newtype {
+            type Err = <$hash as ::std::str::FromStr>::Err;
+            fn from_str(s: &str) -> ::std::result::Result<$newtype, Self::Err> {
+                s.parse().map($newtype)
```
dr-orlovsky: I usually used `$newtype::from_inner(s.parse())`. `s.parse().map($newtype)` looks better, but I assume it can't be used from outside of this mod (since the member is private)?

dr-orlovsky: Let me re-phrase the question. What is the best API for a wrapper types on converting from the wrapped type?

Outside of hash-types case, this have a plenty of other cases, which seems to be quite a generic to rust-language itself. Over the last week I've made >10 types outside of bitcoin-related hash types, and I would like to hear thoughts what would be the most idiomatic API for this - outside of the scope of this PR.

dr-orlovsky: Here is my take on wrapper's idea, with as less macros being involved as possible: https://github.com/LNP-BP/rust-lnpbp/blob/master/src/common/wrapper.rs

elichai:
The CI is failing because of the stupid emscripten. @apoelstra I think you're the only one who can bypass this
(should we make emscripten an optional test that won't fail travis?)

src/lib.rs
stevenroose: Yeah actually I was thinking of using a more explicit method. Like `Ok($newtype(s.parse()?))` But I think in the end it's all the same. Perhaps even `Ok($newtype($hash::from_str(s)?))` would be the best.

PR #79, opened 2020-04-08
Add a hashed fn to EngineTrait

elichai:
This is useful for places like:  https://github.com/rust-bitcoin/rust-bitcoin/pull/422
When encoding into a vec we do a sanity assert that the lengths match https://github.com/rust-bitcoin/rust-bitcoin/blob/master/src/consensus/encode.rs#L154

this will allow us to do it also when using a hasher.

I'm open to suggestions about the function name because technically not all the bytes were "hashed" yet. but I might be bikeshedding it to much.

### DIFF
--- a/src/hmac.rs
+++ b/src/hmac.rs
@@ -107,7 +107,11 @@ impl<T: HashTrait> EngineTrait for HmacEngine<T> {
     }
 
     const BLOCK_SIZE: usize = T::Engine::BLOCK_SIZE;
-    
+
+    fn n_bytes_hashed(&self) -> usize {
+        self.iengine.n_bytes_hashed()
+    }
+
     fn input(&mut self, buf: &[u8]) {
         self.iengine.input(buf)
     }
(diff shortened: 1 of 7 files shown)


stevenroose:
Since the field is called length, how about the traditional function name `len()`? Or does that suggest too much that the engine is a storing bytes internally.

elichai:
`len()` is really used to express actual length, so I think it's confusing here.
what's `nb_hashes_bytes` stands for?

stevenroose:
Sorry that's a typo. Should be `nb_hashed_bytes`. `nb` is quite common for "number".

apoelstra:
I would be okay with `n_bytes_hashed`.

Can you also rebase (I think we've fixed Travis)

elichai:
Actually my use-case is no longer relevant, so if no one else will respond here with a usecase I think I'll just close this PR

stevenroose:
Since it's just exposing a variable that all hashes already keep, it really doesn't hurt to have. So if renamed to `n_bytes_hashed`, I'd just merge it in, just in case anyone else would want it.

Unless we think that we might have hashes in the future that won't keep this state inherently, causing them to have to add an explicit counter just for this.

PR #81, opened 2020-04-24
Make it easy to access the contents of any hash (newtype)

thomaseizinger:
Hashes can be represented as fixed length arrays. That is in fact
what they are internally.
There is value in using Rust's visibility rules to enforce that
no every byte array is a valid hash.
However, every hash is certainly a valid byte array, hence it
should be easy to access their contents as such.

We add a `From` impl for any hash newtype for the inner byte array.
This makes it possible to pass hash-newtypes the way they are to
any function that needs a byte-array.

We also add a `From` impl for any inner hash type that allows us
to convert to the inner representation.

Both of these changes are especially useful from a broader ecosystem
perspective:

Because `From` and `Into` are part of the standard library, any
application or crate depends on them and can write functions that
support a form of method overloading by accepting anything that is
`Into<[u8; 32]>`.

Implementing these conversion traits on our types makes them
compatible with this pattern which results in less verbose APIs.

### DIFF
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -42,6 +42,7 @@ hex_fmt_impl!(LowerHex, Hash);
 index_impl!(Hash);
 serde_impl!(Hash, 20);
 borrow_slice_impl!(Hash);
+from_hash_for_inner_impl!(Hash);
 
 impl str::FromStr for Hash {
     type Err = ::hex::Error;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -181,6 +181,12 @@ macro_rules! hash_newtype {
             }
         }
 
+        impl ::std::convert::From<$newtype> for <$hash as $crate::Hash>::Inner {
+            fn from(hashtype: $newtype) -> <$hash as $crate::Hash>::Inner {
+                hashtype.0.into_inner()
+            }
+        }
+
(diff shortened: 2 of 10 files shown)


stevenroose:
I'm curious what kind of situations you need this for..

thomaseizinger:
> I'm curious what kind of situations you need this for..

If https://github.com/rust-bitcoin/rust-secp256k1/pull/211 is not accepted, then the `From` abstraction might not be as useful because on an individual basis, you can also just use `into_inner`.

One could still argue that it is good use of the standard library to provide these conversion functions if the conversion exists, just to allow a more flexible use of the types :)

apoelstra:
concept ACK. I wonder if this would work now with 1.29 as the MSRV.

thomaseizinger:
> concept ACK. I wonder if this would work now with 1.29 as the MSRV.

I rebased onto latest master and pushed again, let's see what CI thinks.

thomaseizinger:
> concept ACK. I wonder if this would work now with 1.29 as the MSRV.

Nope. We will need Rust 1.41 for this: https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html#relaxed-restrictions-when-implementing-traits

Closing then.

PR #87, opened 2020-09-03
Removing unnecessary `core` and `serde` crate exports

dr-orlovsky:
This was introduced by me some time ago when I was working on hashtype implementation in rust-bitcoin. However it's improper way of doing things, so I've fixed it now. Nevertheless it is a breaking change (since those who was relying on `core` exported by this library will now have to modify their code)


dr-orlovsky:
Closing: again forgot that I've tried to fix this issue once in #73, but failed. Seems like the only way to make it right is to implement derive macros instead of all those proc macros

PR #89, opened 2020-09-03
Renaming Hash/Engine trait to Digest/Engine (implements #16)

dr-orlovsky:
Seems there was an agreement on better trait names, reached in https://github.com/rust-bitcoin/bitcoin_hashes/issues/16, but never implemented. This PR implements it, plus as a separate second and third commits it
- removes unnecessary `use Digest as HashTrait` and `use DigestEngine as EngineTrait`
- renames all cases `MidState` to `Midstate` to match naming of `fn midstate()` (according to Rust naming conventions, https://doc.rust-lang.org/1.0.0/style/style/naming/README.html, we have to either use `type MidState` with `fn mid_state`, if we think it's two different words, or `type Midstate` with `fn midstate()`, if it is a single word)

### DIFF
--- a/fuzz/fuzz_targets/ripemd160.rs
+++ b/fuzz/fuzz_targets/ripemd160.rs
@@ -2,7 +2,7 @@
 extern crate bitcoin_hashes;
 extern crate crypto;
 
-use bitcoin_hashes::Hash;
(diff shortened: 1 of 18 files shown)


dr-orlovsky:
I am planing to close this PR is there is no objections in the next day or two

PR #92, opened 2020-09-15
Remove pub from HexIterator

RCasatta:
struct must be created from new() to check for oddness, otherwise the
unwrap in the iterator impl could be hit


elichai:
I think you're confusing a public struct with public fields?
the struct must be public if we return it from public API, and you can see in the docs that it's not possible to create it without the `new` function: https://docs.rs/bitcoin_hashes/0.9.0/bitcoin_hashes/hex/struct.HexIterator.html

RCasatta:
oh yes, sorry!

PR #93, opened 2020-10-07
Implement Hash::engine() for hash newtypes

stevenroose:
Instead of the default implementation of `Engine::default()`.

This is needed to make newtypes with with `sha256t` tagged hashes for taproot.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -188,6 +188,10 @@ macro_rules! hash_newtype {
             const LEN: usize = <$hash as $crate::Hash>::LEN;
             const DISPLAY_BACKWARD: bool = $reverse;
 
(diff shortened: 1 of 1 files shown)


stevenroose:
Strictly this is a breaking change, but it's only breaking for users of the tagged hashes, which I think there's none. Anyway, I'm patient on this one, because it's only relevant for taproot anyway.

apoelstra:
I disagree that this is a breaking change. There is a visible change in the behavior of the `engine()` method on these hashes, but the old behavior was clearly wrong (and anyone using this API must have just been working around the issue somehow).

apoelstra (approved):

PR #94, opened 2020-10-07
Add sha256t_hash_newtype macro to create tagged hash newtypes

stevenroose:


src/sha256t.rs
```diff
@@ -78,6 +78,28 @@ impl<T: Tag> HashTrait for Hash<T> {
...
 
+/// Macro used to define a newtype tagged hash.
+/// It creates two public types:
+/// - a sha246t::Tag struct,
+/// - a sha256t::Hash type alias.
+#[macro_export]
+macro_rules! sha256t_hash_newtype {
+    ($newtype:ident, $tag:ident, $midstate:ident, $midstate_len:expr, $docs:meta, $reverse:expr) => {
+        /// The tag used for [$newtype].
+        #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Default, Hash)]
```
dr-orlovsky, line 89: What about `Debug`?

apoelstra (approved):
utACK

src/sha256t.rs
```diff
@@ -78,6 +78,28 @@ impl<T: Tag> HashTrait for Hash<T> {
     }
 }
 
+/// Macro used to define a newtype tagged hash.
+/// It creates two public types:
+/// - a sha246t::Tag struct,
+/// - a sha256t::Hash type alias.
+#[macro_export]
+macro_rules! sha256t_hash_newtype {
```
dr-orlovsky, line 86: May be `sha2561_newtype`, since `sha` already implies that it is a hash?

apoelstra: I think the name is for symmetry with `hash_newtype`

src/sha256t.rs
stevenroose: added

src/sha256t.rs
stevenroose: could be a tag... :) idk I think it's good to be inline with `hash_newtype`?

dr-orlovsky (approved):

apoelstra (approved):
re-utACK

dr-orlovsky:
confirm my ACK

dr-orlovsky:
Just understood that this macro uses `hash_newtype!`, which implements `Display` and `FromStr` defaulting to reading/creating hexadecimal representation. However for certain tagged hashes types I use Bech32 encoding, and it's impossible to change that with this macro...

stevenroose:
Why do you use bech32? Hmm yeah well the reason I had these in rust-bitcoin in a taproot module was partly to make sure this wouldn't grow larger than just taproot usage. But still I think as long as there is no Bitcoin usage of tagged hashes that serializes differently, I think it's fine here.

It's really not super hard to specify an alternative macro in your project to create bech32 tagged hash newtypes, right?

PR #95, opened 2020-10-07
Move most of serde trait impl out of macro

JeremyRubin:
This slightly improves the serde macro code by moving most of the impl macro code to external traits and wrappers.

This should have no functional change, but it improves the type safety as the SerdeHash trait is declared  in one place, type checked, and monomorphized where needed.

### DIFF
--- a/src/serde_macros.rs
+++ b/src/serde_macros.rs
@@ -1,86 +1,123 @@
+#[cfg(feature = "serde")]
+pub mod serde_details {
+    use Error;
+
+    use std::marker::PhantomData;
+    struct HexVisitor<ValueT>(PhantomData<ValueT>);
+    use serde::{de, Serializer, Deserializer};
+    use hex;
+
+    impl<'de, ValueT> de::Visitor<'de> for HexVisitor<ValueT>
+    where
+        ValueT: hex::FromHex,
+    {
+        type Value = ValueT;
+
+        fn expecting(&self, formatter: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+            formatter.write_str("an ASCII hex string")
+        }
+
+        fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            if let Ok(hex) = ::std::str::from_utf8(v) {
+                Self::Value::from_hex(hex).map_err(E::custom)
+            } else {
+                return Err(E::invalid_value(
+                    de::Unexpected::Bytes(v),
+                    &self,
+                ));
+            }
+        }
+
+        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            Self::Value::from_hex(v).map_err(E::custom)
+        }
+    }
+
+    struct BytesVisitor<ValueT>(PhantomData<ValueT>);
+
+    impl<'de, ValueT> de::Visitor<'de> for BytesVisitor<ValueT>
+    where ValueT : SerdeHash {
+        type Value = ValueT;
+
+        fn expecting(&self, formatter: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+            formatter.write_str("a bytestring")
+        }
+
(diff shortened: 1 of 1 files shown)


src/serde_macros.rs
```diff
@@ -1,86 +1,124 @@
...
+
+    #[macro_export]
+    /// Implements `Serialize` and `Deserialize` for a type `$t` which
+    /// represents a newtype over a byte-slice over length `$len`
+    macro_rules! serde_impl(
+
+        ($t:ident, $len:expr) => (
+            impl $crate::serde_macros::serde_details::SerdeHash for $t {
+                const N : usize = $len;
+                fn from_slice_delegated(sl: &[u8]) -> Result<Self, $crate::Error> {
```
JeremyRubin, line 97: N.B. this is done this way to disambiguate which function is being called, and because from_slice is defined variously in Midstate and in HashTrait so the macro is needed to handle that genericity.

src/serde_macros.rs
```diff
@@ -1,86 +1,124 @@
...
+            } else {
+                d.deserialize_bytes(BytesVisitor::<Self>(PhantomData))
+            }
+        }
+    }
+
+    #[macro_export]
+    /// Implements `Serialize` and `Deserialize` for a type `$t` which
+    /// represents a newtype over a byte-slice over length `$len`
+    macro_rules! serde_impl(
```
stevenroose, line 92: It seems that now the macro is put into the `serde_details` crate, which seems like it's not supposed.

src/serde_macros.rs
```diff
@@ -1,86 +1,124 @@
...
+                fn serialize<S: $crate::serde::Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+                    $crate::serde_macros::serde_details::SerdeHash::serialize(self, s)
+
+                }
+            }
+
+            impl<'de> $crate::serde::Deserialize<'de> for $t {
+                fn deserialize<D: $crate::serde::Deserializer<'de>>(d: D) -> Result<$t, D::Error> {
+                    $crate::serde_macros::serde_details::SerdeHash::deserialize(d)
+
```
stevenroose, line 112: Both here and in `Serialize` you have some whitespace at the end of the method.

src/serde_macros.rs
```diff
@@ -1,86 +1,124 @@
...
 
 /// Does an "empty" serde implementation for the configuration without serde feature
 #[macro_export]
-#[cfg(not(feature="serde"))]
+#[cfg(not(feature = "serde"))]
 macro_rules! serde_impl(
-    ($t:ident, $len:expr) => ()
-);
+        ($t:ident, $len:expr) => ()
+    );
```
stevenroose, line 124: These seemed to be indented without needing to be.

src/serde_macros.rs
```diff
@@ -1,86 +1,124 @@
+#[macro_use]
```
stevenroose, line 1: So this is probably no longer needed once you move the macro out of this submodule.

src/serde_macros.rs
```diff
@@ -1,86 +1,124 @@
...
+        where
+            E: crate::serde::de::Error,
+        {
+            SerdeHash::from_slice_delegated(v).map_err(|_| {
+                // from_slice only errors on incorrect length
+                E::invalid_length(v.len(), &stringify!(N))
+            })
+        }
+    }
+    pub trait SerdeHash
```
stevenroose, line 61: new line here

stevenroose (dismissed):
Looks good, just a few style nits and then moving the macro out of the submodile.

src/serde_macros.rs
```diff
@@ -1,86 +1,121 @@
+#[cfg(feature = "serde")]
+pub mod serde_details {
+    use Error;
+
+    use std::marker::PhantomData;
+    struct HexVisitor<ValueT>(PhantomData<ValueT>);
+
+    impl<'de, ValueT> crate::serde::de::Visitor<'de> for HexVisitor<ValueT>
```
apoelstra, line 8: I'd prefer if you put a `use serde::{self, de, ser}` at the top of this module

JeremyRubin:
Done.

apoelstra (approved):
utACK

PR #97, opened 2020-10-08
Add Optional JSONSchema Support for Hash Types

JeremyRubin:
This patch adds functionality to optionally be able to generate json schema types.

The generated schemas are not 'complete' verification rules (as they don't know about things like length), but they can be useful for describing an expected return type from e.g. a web client. Future work can add a custom serializer (no great support for declaring this like serde attrs yet) which adds a length validation https://graham.cool/schemars/examples/7-custom_serialization/ and regex to ensure that all fields are hex hashes of appropriate length.

Schemars in particular is a relatively simple library (I've looked it over casually) and has only one new dependency compared to the rust bitcoin crate (dyn-clone).

While this is a WIP, it can be merged. But I'd love to get conceptual feedback here as the plan would be to introduce similar patches across the entire rust-bitcoin project so that any type (transactions, witnesses, blocks, etc) can get automatically generated schemas.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,14 @@ fuzztarget = [] # used by other rust-bitcoin projects to make hashes almost-noop
 
 [dev-dependencies]
 serde_test = "1.0"
+serde_json = "1.0"
+jsonschema-valid = "0.4.0"
 
 [dependencies]
+
+[dependencies.schemars]
+version = "0.8.0"
+optional = true
 
 [dependencies.serde]
 version = "1.0"
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -28,7 +28,13 @@ use Error;
 
 /// Output of the Bitcoin HASH160 hash function
 #[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
-pub struct Hash([u8; 20]);
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct Hash(
+    #[cfg_attr(feature = "schemars", schemars(schema_with="crate::util::json_hex_string::len_20"))]
+    [u8; 20]
+);
+
+
 
 hex_fmt_impl!(Debug, Hash);
 hex_fmt_impl!(Display, Hash);
@@ -145,6 +151,7 @@ mod tests {
     #[cfg(feature="serde")]
     #[test]
     fn ripemd_serde() {
+
         use serde_test::{Configure, Token, assert_tokens};
 
         static HASH_BYTES: [u8; 20] = [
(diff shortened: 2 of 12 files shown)


src/hash160.rs
```diff
@@ -159,6 +165,25 @@ mod tests {
         assert_tokens(&hash.compact(), &[Token::BorrowedBytes(&HASH_BYTES[..])]);
         assert_tokens(&hash.readable(), &[Token::Str("132072df690933835eb8b6ad0b77e7b6f14acad7")]);
     }
+
+    #[cfg(feature="json")]
+    #[cfg(feature="serde")]
```
apoelstra, line 170: Can you use `#[cfg(all(feature="json",feature="serde"))]` rather than doubling up `cfg` lines?

sgeisler:
I really like the concept! This could be very useful for automatically creating API documentation.

src/hash160.rs
JeremyRubin: fixed

JeremyRubin:
Fixed Andrew's Nit.

I also added an implementation for a custom hex string type, so now a generated schema for e.g. sha512 looks like:
```
{"$schema":"http://json-schema.org/draft-07/schema#","description":"Output of the SHA256 hash function","maxLength":128,"minLength":128,"pattern":"[0-9a-fA-F]+","title":"Hash","type":"string"}
```

src/hash160.rs
```diff
@@ -28,7 +28,12 @@ use Error;
 
 /// Output of the Bitcoin HASH160 hash function
 #[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
-pub struct Hash([u8; 20]);
+#[cfg_attr(feature="json", derive(schemars::JsonSchema))]
+pub struct Hash(
+    #[cfg_attr(feature="json", schemars(schema_with="crate::util::json_hex_string::len_20"))]
+    [u8; 20]);
```
stevenroose, line 34: For all of these types, could you close the struct declaration on a new line?
```rust
#[cfg_attr(feature="json", derive(schemars::JsonSchema))]
pub struct Hash(
    #[cfg_attr(feature="json", schemars(schema_with="crate::util::json_hex_string::len_20"))]
    [u8; 20]
);
```

src/hash160.rs
```diff
@@ -28,7 +28,12 @@ use Error;
 
 /// Output of the Bitcoin HASH160 hash function
 #[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
-pub struct Hash([u8; 20]);
+#[cfg_attr(feature="json", derive(schemars::JsonSchema))]
```
stevenroose, line 31: nit: We use `feature="something"` and `feature = "something"` both. Could we add spaces here to eventually migrate to give all of them spaces?

Cargo.toml
```diff
@@ -18,12 +18,19 @@ std = []
 serde-std = ["serde/std"]
 unstable = []  # for benchmarking
 fuzztarget = [] # used by other rust-bitcoin projects to make hashes almost-noops, DON'T USE THIS
+json = ['schemars']
```
stevenroose, line 21: Hmm this might confuse users that they need this feature to have serde serialization for json. Could we either just use the `schemars` feature without adding an explicit feature for it or use something more specific like `json-schema`?

src/util.rs
```diff
@@ -209,6 +209,26 @@ define_slice_to_le!(slice_to_u64_le, u64);
...
+#[cfg(feature="json")]
+pub mod json_hex_string {
+    use schemars::schema::{Schema, SchemaObject};
+    use schemars::{gen::SchemaGenerator, JsonSchema};
+    macro_rules! define_custom_hex {
+        ($name:ident, $len:expr) => {
+
+            pub fn $name(gen: &mut SchemaGenerator) -> Schema {
+                let mut schema: SchemaObject = <String>::json_schema(gen).into();
+                schema.string = Some(Box::new(schemars::schema::StringValidation{max_length: Some($len*2), min_length: Some($len*2), pattern: Some("[0-9a-fA-F]+".to_owned())}));
```
stevenroose, line 221: Can you break this struct declaration up into multiple lines?

stevenroose (changes requested):

src/hash160.rs
JeremyRubin: I'll fix this in my uses, but we don't consistently use either across the code as is.

JeremyRubin:
Fixed most comments.

I did not fix https://github.com/rust-bitcoin/bitcoin_hashes/pull/97#discussion_r502830800. I don't really have an opinion on this, other than that for serde we do use-serde, so I figured that the pattern is not to use the dep name, which is also better if we later add any additional functionality around jsons (not that I have any such plans).

I'm fine with naming it whatever.

Cargo.toml
apoelstra: I agree, I don't understand at all what the point of renaming the feature is

apoelstra:
Needs rebase. Also, please remove the feature rename, it is not true that "the pattern is not to use the dep name", it is only true that sometimes (old versions of) Cargo forces us to rename features.

JeremyRubin:
rebased and renamed

apoelstra (approved):
Looks fine to me. I second Tim's comments elsewhere about not wanting mixed case, but given that json does not even let you precisely specify numbers, I'm not too shaken up by this. If we have more users for this than just Jeremy (and perhaps Blockstream, we have some json services that could use schemas..) we can revisit these sorts of design decisions.

PR #100, opened 2020-10-16
fix serde macro visibility issue, release 0.9.2

apoelstra:
Our last release broke the build for rust-bitcoin iand anything else using the serde impls in this crate) because it moved internals of the serde macros into the non-public `serde_macros` module. This PR exposes the module and bumps the version to fix downstream users.


elichai (approved):
Ops, ACK, let's make sure it works and publish+yank the release ASAP

PR #102, opened 2020-10-18
more serde_macro fixes; bump version to 0.9.3

apoelstra:
Still having rust-bitcoin build failures, this time with rustc 1.29.0, due to some module path import rule changes. Cleaned these up a bit; also replaced `std` by `core` everywhere; I don't know if we support serde+no_std but it's good to be consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 
+# 0.9.2 - 2020-10-18
+
+* Fix rustc 1.29.0 downstream issues with serde macros
+
 # 0.9.2 - 2020-10-16
 
 * Fix visibility issue with serde macros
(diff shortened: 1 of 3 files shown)


elichai (approved):
We should really have better tests :/

apoelstra:
Yeah, I guess for every crate that has macros we ought to have a test/example which uses them. It's a bit of a weird thing, the macro worked within this crate but not outside it, which is a unique-to-macros problem.

apoelstra:
Yeah, I guess for every crate that has macros we ought to have a test/example which uses them. It's a bit of a weird thing, the macro worked within this crate but not outside it, which is a unique-to-macros problem.

JeremyRubin:
it's also particularly funny in this case because the patch in question was designed to make the macro itself have to do less at the expansion site :slaps head:.

PR #103, opened 2020-10-22
Add Hmac::from_inner_engines constructor

stevenroose:
I'm using this to implement a no-std pbkdf2 implementeation for the
bip39 crate.

Should we have an "only use this if you know what you're doing" warning of is that implied?


apoelstra (approved):
utACK.

I think "don't use this unless you're an expert" is implied. I'm not sure a non-expert would even understand what it is that this function does :).

PR #106, opened 2020-10-31
Reformat the manifest dependency section

stevenroose:
The `[dependency.<crate>]` format is not common and is unclear IMO.


elichai (approved):
ACK

apoelstra (approved):
sure

PR #107, opened 2020-11-28
Fix no_std

justinmoon:
Use `core::convert` instead of `std::convert`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,9 @@
 #![cfg_attr(all(test, feature = "unstable"), feature(test))]
 #[cfg(all(test, feature = "unstable"))] extern crate test;
 
+#[macro_use] pub extern crate alloc;
+#[cfg(feature="bare-io")] extern crate bare_io;
+
 #[cfg(any(test, feature="std"))] pub extern crate core;
 #[cfg(feature="serde")] pub extern crate serde;
 #[cfg(all(test,feature="serde"))] extern crate serde_test;
@@ -45,7 +48,7 @@
 
 #[macro_use] mod util;
 #[macro_use] pub mod serde_macros;
-#[cfg(any(test, feature = "std"))] mod std_impls;
+pub mod impls;
 pub mod error;
 pub mod hex;
 pub mod hash160;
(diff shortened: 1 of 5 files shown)


apoelstra:
Why isn't CI catching this?

I will PR to switch to github actions, and redo our scripts and make them work. Unfortunately the build is currently failing in a few ways so it may be a bit of work.

elichai:
> Why isn't CI catching this?

Because we only use this macro in tests, where `std` is enabled anyway.
we should use this macro to create a private type that we don't use or export, just so it will test this outside of tests
(ie. `mod private {hash_newtype!(__testType, ::sha256d::Hash, 32, doc="test")}`)

elichai:
small note, alloc only and no-std are distinctly different

justinmoon:
Converting this to draft for now. I'm starting to add other stuff to get rust-bitcoin working on real MCUs.

apoelstra:
can you rebase on current master? In #109 we did a fair bit of code cleanup including some nostd fixes.

RCasatta:
I see bare-io is now called core2 and from the badge, it looks it requires 1.47+

src/lib.rs
```diff
@@ -45,7 +48,7 @@
 
 #[macro_use] mod util;
 #[macro_use] pub mod serde_macros;
-#[cfg(any(test, feature = "std"))] mod std_impls;
+pub mod impls;
```
dr-orlovsky, line 51: Why we need this to be `pub`? It does not define any types to export

dr-orlovsky (commented):
What is the current status on this taking into account all other work on `no_std` with bringing `core2` support?

apoelstra:
cc @justinmoon I think we've decided to go with core2 rather than bare-io, see recently-merged #107 and #130.

Should we close this PR or would you like us to revisit it?

TheBlueMatt:
bare-io was renamed core2. core2 is just the new version (though still beta, AFAIU). I think this can be closed.

apoelstra:
Closing. Feel free to reopen @justinmoon if you'd like us to revisit this.

PR #117, opened 2021-04-05
Reintroduce Schemars with MSRV compatible deps

JeremyRubin:
Reverts 4665e8ddb11cdc75f2cb05432e7a7b4fdf29bafd to reintroduce https://github.com/rust-bitcoin/bitcoin_hashes/pull/97

Should be MSRV compatible.


stevenroose:
I'm getting

```
$ ./test-msrv.sh 
error: failed to parse manifest at `/home/steven/code/rust/bitcoin_hashes/Cargo.toml`

Caused by:
  feature `rename-dependency` is required

this Cargo does not support nightly features, but if you
switch to nightly channel you can add
`cargo-features = ["rename-dependency"]` to enable this feature
```

My script simply does 
```sh
#!/bin/sh

set -e

MSRV="1.29.0"

CMD="rustup run ${MSRV}"

rm -f Cargo.lock
$CMD cargo generate-lockfile
$CMD cargo update --package "serde_json" --precise "1.0.39"

$CMD cargo test
$CMD cargo test --features "serde"
```

JeremyRubin:
yeah was just fixing that; sorry for the line noise here...

JeremyRubin:
I think this probably won't work for 1.29 as I didn't realize the cargo edition flag would be a blocker.

if we bump to 1.39 or something given the mrustc update, then groovy, otherwise will extract tests into a new crate.

apoelstra:
Closing. Can re-open if we revisit MSRV.

JeremyRubin:
@apoelstra are you not ok with the tests in separate crate version? The schemars patches themselves worth with current MSRV I think?

apoelstra:
Ok, I'll leave this open -- will review if you can get CI to pass :)

JeremyRubin:
Introduced an extended tests directory of crates which only runs on stable or beta, not 1.29. This removes jsonschema-valid dependencies, and makes bitcoin_hashes only require Schemars.

apoelstra (approved):
ack 133e97df611f116b918729a08b624bf7c8c38e18

PR #131, opened 2021-06-27
core2 0.3.0 was released

devrandom:
Also see https://github.com/rust-bitcoin/rust-bitcoin/pull/603


TheBlueMatt (approved):

sgeisler (approved):
utACK 5c6f313ab1e224c7e547c8d37799a399f2af1a0b

But note that this is breaking. So we'll need a major version bump here and also in rust-bitcoin (which is fortunately planned anyway). If anything non-breaking should go int hashes we should hold off merging this one.

apoelstra:
I think let's just go for it.

PR #133, opened 2021-07-15
Implement `Write` for `HmacEngine`

elsirion:
For some reason that wasn't implemented yet.


apoelstra (approved):
ACK f1084bf63b518f76d1f77abe7709ceabd4ef65ee

Not sure why CI is not running. Passes locally for me.

PR #137, opened 2021-10-28
Allow unknown lints

tcharding:
We support many versions of rustc by design, this puts us in the following predicament:

1. Linting with a newer version of clippy throws a warning for a newly introduced default lint.
2. In order to allow the lint we must use the lint by name but this lint does not exist in earlier versions of clippy, so we get a warning `unknown lint` when linting with the earlier version.

We want to have our cake and eat it too, we can do this by allowing unknown lints. Doing so enables (1) while preventing (2).

If this PR is deemed a good solution we can do the same in `rust-secp256k1` and `rust-bitcoin`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,9 @@
 //! are needed to display hashes anway.
 //!
 
+// Unknown lint warnings occur when linting with an old version of clippy and
+// hitting code that allows a lint that exists only in newer versions of clippy.
+#![allow(unknown_lints)]
 // Coding conventions
 #![deny(non_upper_case_globals)]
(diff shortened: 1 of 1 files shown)


apoelstra:
concept ACK from me.

IIRC in Rust 1.22 we would get actual errors when referencing unknown lints ... I guess 1.29 is smarter than that?

tcharding:
Based off comments in the same PR as this on [rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/pull/688#issuecomment-961878706), this PR is not useful. Please excuse the noise.

PR #141, opened 2022-01-20
Add a disabled rustfmt.toml

tcharding:
We do not currently use `rustfmt`, this is a nuisance for devs who routinely work on code bases that do use rustfmt who often have their editors set to format on save. We can make the life of such devs much better by explicitly disabling formatting using `rustfmt.toml`.

ref: https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#disable_all_formatting

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,1 @@
+disable_all_formatting = true


apoelstra (approved):
ACK bd59192704706c9fdec76bc03890664b2404cecb

PR #142, opened 2022-01-24
Use SliceIndex trait to implement Index

tcharding:
Since we bumped to Rust 1.29 it has not been necessary to implement `Index` manually, we can get it by way of the `SliceIndex` trait.

Implement `Index` using `SliceIndex`. Keep the declarative macro so we only have to write the `SliceIndex` implementation once.

This is in the same vein as https://github.com/rust-bitcoin/rust-bitcoin/pull/805 but goes about it in a slightly different manner, we should probably be uniform across both crates. I'm unsure which is the best option?

### DIFF
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -20,6 +20,8 @@
 //! # HASH160 (SHA256 then RIPEMD160)
 
 use core::str;
+use core::ops::Index;
+use core::slice::SliceIndex;
 
 use sha256;
 use ripemd160;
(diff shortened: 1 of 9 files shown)


apoelstra:
I think, now that we have `SliceIndex`, we could just inline this and drop the macro entirely. It's a bit duplicative but not nearly as bad as the 1.0-era "implement four different `Index` traits" code.

apoelstra (approved):
ACK a94b232ab3d8260346f8cbbef2ca8b1aea593b87

apoelstra:
BTW - this is a breaking change because we exported this macro and downstream users may have been using it :(

tcharding:
Do we need to add a label to this to remind us we broke the API?

apoelstra:
I think it'll be okay -- we'll go over all the merged PRs when making the changelog for the next release.

PR #143, opened 2022-01-25
Remove redundant slicing

tcharding:
Clippy emits:

warning: redundant slicing of the whole range

As suggested use `key` directly instead of the slice.


apoelstra (approved):
ACK d63557fe5c5cc6a9566e1715806edf709189e67b

PR #147, opened 2022-01-27
Improve feature coverage in CI test suite

tcharding:
Improve feature testing by adding all the features to the test matrix and enabling testing with `--no-default-features`.

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -28,16 +28,16 @@ if [ "$DO_FEATURE_MATRIX" = true ]; then
 
     # All features
     cargo build --all --no-default-features --features="$FEATURES"
-    cargo test --all --features="$FEATURES"
+    cargo test --all --no-default-features --features="$FEATURES"
     # Single features
     for feature in ${FEATURES}
     do
         cargo build --all --no-default-features --features="$feature"
-        cargo test --all --features="$feature"
+        cargo test --all --no-default-features --features="$feature"
 		# All combos of two features
 		for featuretwo in ${FEATURES}; do
 			cargo build --all --no-default-features --features="$feature $featuretwo"
-			cargo test --all --features="$feature $featuretwo"
+			cargo test --all --no-default-features --features="$feature $featuretwo"
 		done
     done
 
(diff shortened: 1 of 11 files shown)


apoelstra:
I'm hesitant about the `have_allocator` feature ... it is tempting, and possibly gives extreme power users a way to hack past our "real" feature matrix, but it can be simulated using `#[any(...)]` and `#[all(...)]` within the code itself, which is a bit nicer from a user perspective (even though it can be harder to keep in sync rom our persective)

tcharding:
No worries, I'll pull that change out of the PR and re-spin. Thanks for looking.

contrib/test.sh
```diff
@@ -41,6 +42,9 @@ if [ "$DO_FEATURE_MATRIX" = true ]; then
 		done
     done
 
+    # Other combos
+    # TODO: Add this test once we bump MSRV past 1.29
+    # cargo test --all --no-default-features --features="std,schemars"
```
apoelstra, line 47: In af992d7e9abd4acd0fa0ec1b622e35498d1f6971:

Did you intend to comment out this line? From your commit message it sounds like you *do* intend to test std and schemars.

We definitely need schemars to be tested in CI somehow. Historically we have had breakage from this dependency that went undetected by CI. (This is why, incidentally, I worte all my own independent CI scripts, and also why schemars is not a dependency in any other rust-bitcoin crate.)

tcharding: Shit, I've been a bit sloppy lately fixing commit messages when re-basing. I'm still struggling big time with Rust 1.29, I have some shell functions to build with 1.29 but I often forget to use them because they are not well tied in with my editor ... so I'm constantly re-basing and getting punished by CI. You might have noticed all the force pushes in my PRs :) I'm being lazy, we are so close to bumping to 1.41

We have `extended_tests` which test `bitcoin_hashes` with `serde`, `schemars`, and `std` enabled. The `schemars` feature doesn't build on its own at the moment `cargo check --no-default-features --features=schemars` fails. I didn't investigate why I just added this comment to remind us.

I added this commented out line with the MSRV comment since I've seen that done in a few places and have started doing it. Now as I write this message I think it would be better to add all these to the edition 2018 tracking issue instead of leaving it in the code. 

Will fix the commit log.

apoelstra: Ok, good observation that we'll be bumping to 1.41.0 soon, and at that point we can put schemars into the normal CI pipeline.

apoelstra (approved):
ACK 5dc3fd407b61d97ec73bcc0dfc97391869550f06

No user-visible changes, good to merge.

PR #148, opened 2022-02-02
Add all_zeros method to Hash trait

### LINKED ISSUE #139: `impl Default for sha256::Hash`  is weird
I find all zeroes as the default value weird. Why it should be considered "default hash"? There's not even a known input to all zeroes hash. If there even exists something like "default hash" I'd expect it to be the hash of empty input.

This also looks like a footgun to me. We could have a constant called `ALL_ZEROES` for initialization purposes, I just wouldn't call it "default".

If there's some literature referring to all zeroes as the default hash value I'd be happy to learn but so far I didn't find anything like it.

tcharding:
Currently we use `Default::default()` to get an all zeros hash. This is confusing because an all zeroes hash does not have a real meaning since there is no known input to create it. The all zeros hash is, however used by the Bitcoin network to signify certain things (e.g. the previous blockhash in the genesis block).

Remove the `Default` implementation from the `Hash` trait and add a `all_zeros` method instead. Remove the `Default` derive from any hash type that has it.

To see the patch to `rust-bitcoin` that this PR imposes, check out [this branch](https://github.com/tcharding/rust-bitcoin/tree/bitcoin_hashes-all-zeros).

Resolves #139

### DIFF
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -30,7 +30,7 @@ use Hash as HashTrait;
 use Error;
 
 /// Output of the Bitcoin HASH160 hash function.
-#[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[repr(transparent)]
 pub struct Hash(
@@ -99,6 +99,10 @@ impl HashTrait for Hash {
 
     fn from_inner(inner: Self::Inner) -> Self {
         Hash(inner)
+    }
+
+    fn all_zeros() -> Self {
+        Hash([0x00; 20])
     }
 }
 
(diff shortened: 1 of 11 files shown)


apoelstra (approved):
ACK 51be9bd8261a6a66070db6a488227eba25cdc148

This is a breaking change and requires a major version bump.

tcharding:
> This is a breaking change and requires a major version bump.

Whats the policy for changes like this across the rust-bitcoin stack? Should this PR have included a change to bump the version number or do we rely on you to work this out when you do the release?

apoelstra:
@tcharding we don't have an explicit policy ... but yes, "we rely on me to work it out" is usually what we do.

Sometimes people want a minor version bump for some specific feature, in which case I ask them to put the bump into the PR ... but for major ones I don't think we ever do that.

tcharding:
Friendly bump please review crew.

apoelstra:
Let's do it. We can bring this into the MSRV-bumping release, this crate is simple enough that I don't worry about breaking too much at once.

Issue #124, opened 2021-04-29
Change `FromHex` trait to have associated type Error instead of hex::Error


sanket1729:
The `hex::Error` only has limited enums that make it difficult to implement the trait outside of this crate. I think it would be cleaner to have an associated Error type like `FromStr`. 

This will be a breaking change that will impact all implementations of this trait. Looking for some feedback before implementation.

apoelstra:
I think for the specific issue you're running into (wanting to reject an all-zeroes value) it might make more sense to just add a `OutOfRange` or `InvalidValue` variant to `hex::Error`.

Using an associated error type puts more of a burden downstream to get all the types to line up, and we'd need to require some sort of trait/compatibilty with our existing error returns.

stevenroose:
Yeah or a `Custom(Box<fmt::Display>)` or `Custom(&'static str)` or something, like serde has.

Kixunil:
Could also do something like this:

```rust
use core::convert::Infallible;

trait FromHex: FromHexRestricted<InvalidValueError=Infallible> {
// same as current FromHex
}

// please suggest a better name
pub enum MaybeInvalidValue<E> {
    InvalidValue(E),
    // Could also duplicate the single variant to make some things easier
    Encoding(HexError),
}

trait FromHexRestricted {
    type InvalidValueError;

    // same API as FromHex, except the error is MaybeInvalidValue<Self::InvalidValueError>
}

impl<T: FromHex> FromHexRestricted for T {
    type Error = Infallible;

    // just call the other trait and map error to MaybeInvalidValue::Encoding
}
```

* Fully backwards-compatible thanks to implications.
* Crates can loosen the trait bounds by replacing `FromHex` with `FromHexRestricted` at any time, old code still works because of implication, the errors just need to be handled differently after change
* Maybe there are potential extension traits to make errors easier

Or a lighter version that breaks exhaustive matching, is a bit annoying regarding conversions but involves one less type:

```rust
enum Error<InvalidValue = Infallible> {
    InvalidLength(usize, usize), // or if we're already breaking this InvalidLength { expected: usize, provided: usize, }
    InvalidValue(InvalidValue),
}
```

Kixunil:
I was thinking about this and came to conclusion we should probably remove `FromHex` trait:

* Things that are always parsed as hex should have `FromStr` instead.
* Things that may be parsed as hex but not always are integers and those are not implemented by `bitcoin_hashes`.

This is related to #169

[...]

PR #149, opened 2022-02-02
Add custom error variants

tcharding:
When users of the library attempt to implement `FromHex` they are restricted to returning an error defined by the variants of `hex::Error`. This gives implementors little choice if they are attempting to return an error that is application specific.

This PR is an attempt to implement the idea suggested by @Kixunil in #124 

This PR makes calling the `FromHex` methods a PITA now because callers must resolve the naming conflict every time they use the trait, is this satisfactory or is there a better way? This seems like a standard Rust problem so there should be a standard Rust solution.
```
<T as FromHex>::from_byte_iter()
```

Resolves: #124 

### Note

This PR makes the usage of `FromHex` much more cumbersome, see the unit test changes for an example. Is this satisfactory?

### DIFF
--- a/src/hex.rs
+++ b/src/hex.rs
@@ -26,6 +26,8 @@ use std::io;
 use core2::io;
 
 use core::{fmt, str};
+use core::convert::Infallible;
+
 use crate::Hash;
 
 /// Hex decoding error.
@@ -58,7 +60,10 @@ pub trait ToHex {
 }
 
 /// Trait for objects that can be deserialized from hex strings.
-pub trait FromHex: Sized {
+///
+/// Implement this trait if the `Error` enum is suitable for returning all your errors, if you need
+/// an additional custom error consider implementing `FromHexRestricted`.
+pub trait FromHex: FromHexRestricted<CustomError = Infallible> + Sized {
     /// Produces an object from a byte iterator.
     fn from_byte_iter<I>(iter: I) -> Result<Self, Error>
     where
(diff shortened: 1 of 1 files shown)


Kixunil:
>  I was unable to understand the benefit of the more complicated suggestion made by @Kixunil.

A huge one is not matching against arbitrary strings to check for specific errors and being able to pass more information. In practice this is very important for localization. (I did actually implement localization a long time ago and it was awful because of `io::Error`.) Also being able to signal that `Custom` is actually impossible helps with performance a bit.

Would appreciate next time asking first when you don't understand something I say. I'd be a better use of time.

apoelstra:
> Would appreciate next time asking first when you don't understand something I say. I'd be a better use of time.

In fairness, he did open a PR and made a point of highlighting you (and nobody would merge this without your appearance and your comments being addressed). I don't think the intention here was to ignore or go around you, only to attempt to make some forward progress.

tcharding:
Its all about learning for me, I try to take the action that makes me learn the most. I have a bad tendency to not internalize things when I'm told them but soon as I've made some action (written some code in this instance) then I remember much better the lessons learned. In this case it took barely 20 minutes to hack up a solution. I appreciate both your inputs @apoelstra and @Kixunil, I do also en devour to take actions that don't impose on your time, though in this I can always improve.

tcharding:
So good to have clippy running on CI :)

apoelstra:
Reviewed 90f9ae0b7f5a4314500a1c6f613cfce49b5cb827. Neat! `Infallible` really makes our lives easier.

I think we want to reverse the roles of `FromHex` and `FromHexRestricted`, so that the "default" one is the one that allows custom error types. Otherwise people writing generic code are unlikely to support custom error types, even when they otherwise could.

tcharding:
If we do that we loose the backwards compatibility benefits? If we are going to do that we might as well just add the custom error associated type to the `FromHex` trait and not bother with the other one, right? Implementors can just use `Infallible` themselves as the associated type if they do not need a custom error.

apoelstra:
Ah, I see, I missed that backward compatibility was part of the motivation for this strategy.

Honestly I'm leaning toward just adding an extra variant and breaking compatibility ... this strategy maintains backward-compatibility but also doesn't provide a lot of benefit to users. If I create a type and impl `FromHexRestricted` on it, it's not going to be usable with any existing code that expects a `FromHex`, and I don't expect this situation to improve much in the future since I'd guess most `FromHex`-related code is write-once-never-touch-again code.

tcharding:
Cool, will re-work the PR, converting to draft till its done. Thanks.

tcharding:
> We could add non_exhaustive while we're at it.

Threw an additional patch on https://github.com/rust-bitcoin/bitcoin_hashes/pull/155

tcharding:
I had a play with adding a `Custom(E)` variant, that makes error become `pub enum Error<E>`. Turns out this is a real PITA because of the std/alloc feature. The `E` generic type has to be constrained
```
error[E0207]: the type parameter `E` is not constrained by the impl trait, self type, or predicates
```
Which leads to this horrendous code, or am I missing something?
```rust
impl<E> Foo for Bar {
    fn foo() -> Result<(), Error<
        #[cfg(any(feature = "std", feature = "alloc"))] E: error::Error,
        #[cfg(not(any(feature = "std", feature = "alloc")))] E,
        > {
        ...
    }
}
```
Is there a better way?

apoelstra:
Lol.  Do we actually ever *need* to constrain `E` to be `std::Error`?

tcharding:
It won't work with `source` correctly if not but I guess we could just constrain in with `fmt::Display` and be done with it?

apoelstra:
I'm not sure exactly what you mean, but yes, I think you can restrict it only in the places where it's directly needed.

PR #150, opened 2022-04-26
Build docs using nightly toolchain

tcharding:
In order to get access to the `--cfg docsrs` we need to use the nightly
toolchain. Doing so means the CI docs build mirrors the docsrs docs
build so better tests our docs build prior to release.

PR #151, opened 2022-04-26
Use nightly docs features

tcharding:
After https://github.com/rust-bitcoin/bitcoin_hashes/pull/150 merges we can use the nightly toolchain and get access to the docrs compiler attribute.

Patch 1 and 2 are cleanup in preparation for patch 3 which adds the experimental docrs cfg attribute of form
```
#[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
```
to all non-test code that is feature gated.


apoelstra (approved):
ACK ae03655314bc6a042ad37d0f98bd6f156ba7c141

PR #153, opened 2022-04-27
Add docs to the code create by sha256t_hash_newtype macro

tcharding:
When using the `sha256t_hash_newtype` currently no docs are generated,
this causes lint errors for users who lint for such things.

Add docs to the new type created by the `sha256t_hash_newtype` macro.

(This is identical to how we do it in the duplicate macro of the same
name in `rust-bitcoin`, which can be removed in favour of this one after
this patch is applied.)


apoelstra (approved):
ACK 275adc6c335a4326699cfbd444949e1725864ea1

PR #154, opened 2022-05-25
Remove leading double colons

tcharding:
Leading double colons are a relic of Rust edition 2015, we do not need to use them any more now we are on edition 2018.


apoelstra (approved):
ACK 49ba623367b4f4cec8bef6a284fdf078591be14a

PR #156, opened 2022-06-01
Introduce HexWriter

RCasatta:
This struct create an hex string from a writer, in cases where you have
an encodable object this allows skipping the creation of the intermediate `Vec<u8>`

Useful in rust_bitcoin `serialize_hex` https://github.com/rust-bitcoin/rust-bitcoin/blob/0e82376bf85137494ff220608e80b9c78c1932b5/src/consensus/encode.rs#L149 but needs https://github.com/rust-bitcoin/rust-bitcoin/pull/1027 so that result string size is known ahead of time

```
test hex::benches::bench_to_hex                   ... bench:       2,669 ns/iter (+/- 233)
test hex::benches::bench_to_hex_writer            ... bench:       2,034 ns/iter (+/- 167)
```

### DIFF
--- a/src/hex.rs
+++ b/src/hex.rs
@@ -20,9 +20,9 @@ use crate::alloc::{string::String, vec::Vec};
 #[cfg(feature = "alloc")]
 use crate::alloc::format;
 
-#[cfg(feature = "std")]
+#[cfg(any(test, feature = "std"))]
 use std::io;
-#[cfg(all(not(feature = "std"), feature = "core2"))]
+#[cfg(all(not(test), not(feature = "std"), feature = "core2"))]
 use core2::io;
 
 use core::{fmt, str};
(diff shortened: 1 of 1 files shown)


src/hex.rs
```diff
@@ -219,6 +219,51 @@ impl ToHex for [u8] {
...
+#[cfg(any(test, feature = "std", feature = "alloc"))]
+#[cfg_attr(docsrs, doc(cfg(any(test, feature = "std", feature = "alloc"))))]
+pub struct HexWriter(String);
+
+#[cfg(any(test, feature = "std", feature = "alloc"))]
+#[cfg_attr(docsrs, doc(cfg(any(test, feature = "std", feature = "alloc"))))]
+impl HexWriter {
+    /// Creates a new [`HexWriter`] specify the `capacity` of the inner `String` that will contain
+    /// the hex value
+    pub fn new(capacity: usize) -> Self {
```
apoelstra, line 241: I think `new` should take no parameters and we should have a separate `with_capacity` constructor which takes a capacity.

tcharding:
CI is because you include `test` in the cfg feature gating but `io` is not imported for `test`.

src/hex.rs
```diff
@@ -219,6 +219,51 @@ impl ToHex for [u8] {
...
+/// in an hex String
+///
+/// If you already have the data to be converted in a `Vec<u8>` use [`ToHex`]
+/// but if you have an encodable object, by using this you avoid the
+/// serialization to `Vec<u8>` by going directly to `String`.
+///
+/// Note that to achieve better perfomance than [`ToHex`] the struct must be
+/// created with the right `capacity` of the end hex string result so that the
+/// inner `String` doesn't re-allocate
+#[cfg(any(test, feature = "std", feature = "alloc"))]
```
tcharding, line 232: Some minor grammar fixes if you want them
```
/// A struct implementing [`io::Write`] that converts what's written to it into
/// a hex string.
///
/// If you already have the data to be converted in a `Vec<u8>` use [`ToHex`]
/// but if you have an encodable object, by using this you avoid the
/// serialization to `Vec<u8>` by going directly to `String`.
///
/// Note that to achieve better perfomance than [`ToHex`] the struct must be
/// created with the `capacity` of the final hex string so that the inner
/// `String` doesn't re-allocate.
```

TIL, for words starting with 'h' we use 'an' before them if the 'h' is silent (e.g., an hour) and we use 'a' before them if the 'h' is pronounced (e.g. a hat).

src/hex.rs
```diff
@@ -219,6 +219,51 @@ impl ToHex for [u8] {
...
+#[cfg_attr(docsrs, doc(cfg(any(test, feature = "std", feature = "alloc"))))]
+pub struct HexWriter(String);
+
+#[cfg(any(test, feature = "std", feature = "alloc"))]
+#[cfg_attr(docsrs, doc(cfg(any(test, feature = "std", feature = "alloc"))))]
+impl HexWriter {
+    /// Creates a new [`HexWriter`] specify the `capacity` of the inner `String` that will contain
+    /// the hex value
```
tcharding, line 240: Perhaps:
```
    /// Creates a new [`HexWriter`] with the `capacity` of the inner `String`
    /// that will contain final hex value.
```

src/hex.rs
```diff
@@ -219,6 +219,51 @@ impl ToHex for [u8] {
...
+    pub fn new(capacity: usize) -> Self {
+        HexWriter(String::with_capacity(capacity))
+    }
+
+    /// Get the hex string result
```
tcharding, line 245: Perhaps
```
    /// Returns the resulting hex string.
```

tcharding (commented):
With the import fix, ACK.

src/hex.rs
RCasatta: thanks, will apply

src/hex.rs
RCasatta: thanks, changed.

src/hex.rs
RCasatta: thanks, changed.

src/hex.rs
RCasatta: What mentioned @tcharding is the reason why I avoided having the `new` without parameters. @apoelstra do you think it's okay?

apoelstra (approved):
ACK 7e796285a52a2a38e743474fa3ae94d76d6a1699

apoelstra:
@tcharding are your nits addressed?

tcharding:
Yep, thanks for checking with me!

tcharding (approved):
ACK 7e796285a52a2a38e743474fa3ae94d76d6a1699

PR #158, opened 2022-06-02
Remove lib section from manifest

tcharding:
The lib section we have is the same as the defaults, it is unnecessary.
Now that we are using edition 2018 it makes sense to use the more modern
form and not include an explicit lib section.


apoelstra (approved):
ACK 36d0ea1fbd927198d22b2ab553ffa3093a8a0077

PR #161, opened 2022-06-07
Fix macro usage in macros

tcharding:
Currently we have macros that call other macros without a fully qualified path, this breaks some usage of the macros downstream because macro calls that should be opaque require importing.

Fix macro usage in macros by using `$crate::` fully qualified path.

Done in preparation for removing `extern crate` and `macro_use` in `rust-bitcoin` (since we have edition 2018).

### DIFF
--- a/src/sha256t.rs
+++ b/src/sha256t.rs
@@ -162,7 +162,7 @@ macro_rules! sha256t_hash_newtype {
             }
         }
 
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK aba33d1042b26252ad2f15a92f31a139956c9e8b

apoelstra:
Thanks for chasing this down! Macro usability is really irritating to test.

PR #166, opened 2022-07-13
Create configuration conditional "bench"

tcharding:
As we did in rust-bitcoin [0] create a configuration conditional bench
that we can use to guard bench mark code. This has the benefit of
making our features additive i.e., we can now test with --all-features
with a stable toolchain (currently this fails because of our use of the
test crate).

[0] - rust-bitcoin/rust-bitcoin#1092


apoelstra:
Will wait to nail down the rust-bitcoin version before reviewing/merging this.

tcharding:
I'm unsure right now why CI is failing.

apoelstra:
Looks like the `dyn-clone` dep broke in a minor version bump. This is a dependency of `schemars`. Probably we should just drop schemars again.

cc @JeremyRubin

tcharding:
`dyn-clone` 1.0.8 breaks our build, I've created a patch to pin the transitive dependency, done as an additional option to removing the `scemars` dependency. I have no knowledge or opinion on which is best. 

https://github.com/rust-bitcoin/bitcoin_hashes/pull/167

tcharding:
Changes in force push:
- rebase
- Check against `RUSTUP_TOOLCHAIN` for error message
- Send output from grep to `/dev/null` when setting `NIGHTLY` (should have been done in rust-bitcoin as well)

apoelstra (approved):
ACK a7b7bed5625c309034dada9a34c7867872205eca

PR #167, opened 2022-07-21
Hack around the broken dyn-clone dependency

tcharding:
The current version of `dyn-clone` v1.0.8 breaks our build (uses `Arc::as_ptr` which only became available in Rust 1.45). 

We can pin the dependency to a version before that works. Requires some toml magic akin to what we do with `actual-serde` in `rust-bitcoin`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ edition = "2018"
 [features]
 default = ["std"]
 std = []
+schemars = ["actual-schemars", "dyn-clone"]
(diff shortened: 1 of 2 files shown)


apoelstra:
Cool, I think this minimally irritates both me and @JeremyRubin. Let's do it.

JeremyRubin:
Looks fine to me! Thanks

apoelstra (approved):
ACK 67aa7ac973d4fe6843832aeee9a1f20c62065260

PR #173, opened 2022-09-01
Use stdlib integer/slice conversion methods

tcharding:
Now that we have Rust > 1.32 we can use methods on stdlib integers to convert to and from little and big endian slices.

- Patch 1: preparatory cleanup
- Patch 2: do the work
- Patch 3: optimisation, use `chunks_exact`

### DIFF
--- a/src/ripemd160.rs
+++ b/src/ripemd160.rs
@@ -79,7 +79,7 @@ impl crate::HashEngine for HashEngine {
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[repr(transparent)]
 pub struct Hash(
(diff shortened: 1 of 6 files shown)


apoelstra:
Looks good! Needs rebase.

Sad that `chunks_exact` doesn't let us avoid the `.expect`s when converting from slices to arrays..

apoelstra (approved):
ACK db2530f6ac60817be3e7e08c2fb7a20578215633

tcharding:
> Sad that chunks_exact doesn't let us avoid the .expects when converting from slices to arrays..

Agree, I had a bit more of a play with it and the best I could come up with was adding a local `[u8; 4]` and manually copying into it but that is even more ugly.

apoelstra (approved):
ACK 6f9c0ddab979d1adf0f366047ac2330e856250cb

PR #174, opened 2022-09-01
Implement `Debug` and `Display` using `LowerHex`

tcharding:
Patch 2 adds a unit test that can be moved to before patch 1 to verify this PR does what it claims.

From the commit log of patch 1:

`Display` and `LowerHex` offer an 'alternate' option, hex representations by convention should include `0x` if the alternate format is printed.
    
Currently we implement `Display`, `Debug`, and `LowerHex` all using the same macro. This has two problems, the biggest of which is that we are not correctly handling the alternate format. The second issue is code duplication.
    
Modify the `impl_hex_fmt` macro to implement `LowerHex` (incl. alternate form), then implement `Display` and `Debug` in the macro making calls to `LowerHex`.

### DIFF
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -35,9 +35,7 @@ pub struct Hash(
     [u8; 20]
 );
 
-hex_fmt_impl!(Debug, Hash);
-hex_fmt_impl!(Display, Hash);
-hex_fmt_impl!(LowerHex, Hash);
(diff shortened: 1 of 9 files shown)


apoelstra (approved):
ACK d3da48e42656576f3a6131f3bfc50e1f492763aa

PR #178, opened 2022-09-01
Deduplicate hash types

Kixunil:
There was a lot of code duplicated for all hash types. Most of it
various standard impls. This made it more tedious and error-prone to
modify the impls for all types.

This commit creates two new macros that take care of implementing the
hash types and calls them from each hash module as needed to keep API
and behavior without a change.

This also extends the `serde_impl!` macro to support generics.

This is preparation for upcoming changes.

### DIFF
--- a/src/hash160.rs
+++ b/src/hash160.rs
@@ -26,79 +26,24 @@ use core::slice::SliceIndex;
 
 use crate::{Error, hex, ripemd160, sha256};
 
-/// Output of the Bitcoin HASH160 hash function.
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
-#[repr(transparent)]
-pub struct Hash(
-    #[cfg_attr(feature = "schemars", schemars(schema_with = "crate::util::json_hex_string::len_20"))]
-    [u8; 20]
-);
-
-hex_fmt_impl!(Hash);
-serde_impl!(Hash, 20);
-borrow_slice_impl!(Hash);
-
-impl<I: SliceIndex<[u8]>> Index<I> for Hash {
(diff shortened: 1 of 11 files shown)


apoelstra (approved):
ACK bd2304fdf94ff4515b345707eb9304e7f611151e

apoelstra:
Looks good! Needs rebase.

Would like another ack before merging since this does make the code a fair bit harder to read by macro-izing it, even though it does dramatically reduce repetition and boilerplate.

Kixunil:
The main benefit is updating everything in a single change. Being slightly less readable should be less bad than risking inconsistent API.

apoelstra (approved):
ACK c5bb14c755eafcb3496e7d873bf0dc71e12e01e8

tcharding (approved):
Having spent a good part of yesterday doing the same edits in every hash file I really like this.

I did a few further deduplications and added some consts to assist readability, see branch of the same name [on my tree](https://github.com/tcharding/bitcoin_hashes/tree/deduplicate-hash-types) if you like.

I changed the macro invocation to be
```rust
const DISPLAY_BACKWARDS: bool = false;
const NBITS: usize = 256;

crate::internal_macros::hash_type! {
    "Output of the SHA256 hash function.",
    "crate::util::json_hex_string::len_32"
}
```

Since we already rely on things being in scope I figured why not rely on the consts being there too.

tcharding:
ACK c5bb14c755eafcb3496e7d873bf0dc71e12e01e8

Kixunil:
I find parameters easier to write and more flexible.

tcharding:
No worries

Kixunil:
Test failures seem unrelated.

apoelstra (approved):
ACK 847f2812b6f033e5a152cd277ebc25adf8b88d10

PR #185, opened 2022-09-07
move schemars to separate CI job

apoelstra:
and run it on nightly only

If it breaks again we'll set it to "allow fail".

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -80,6 +80,7 @@ jobs:
           - 1.41.1
           - beta
           - stable
+          - nightly
     steps:
       - name: Checkout Crate
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


tcharding (approved):
ACK d0a77d6f9f4d5f70c306290f15a6a8dbf7575101

This is still going to build the `schemars` feature in other jobs - do we want to limit that to this job also?

apoelstra:
@tcharding good catch. I think for now it's ok. The issue here is that the extra schemars test has a pile of bad dependencies of its own, even beyond schemars itself.

tcharding:
Sweet, cheers.

Kixunil:
Concept ACK, why nightly though? Yes, we may catch and report Rust error which is noble but I don't think we should fail CI if something breaks. Perhaps allow fail on nightly and disallow fail on stable?

apoelstra:
>why nightly though

It seemed least likely to break again. Maybe `stable` would've been a better choice?

Kixunil (approved):
ACK 64c085cde284fad572a1f8429abc4331f61b44e9

Kixunil:
Oh, actually this changes MSRV of `shemars` to `stable`. Would be better to document that.

apoelstra:
There has never been a `MSRV` for schemars, so this is not a change.

Kixunil:
> This library should always compile with any combination of features on **Rust 1.41.1**.

`schemars` is a feature and it being enabled is one of the possible combinations of features.

apoelstra:
:( ok I'll update the text. I don't know how that happened, schemars has repeatedly broken MSRV

tcharding:
I'd hazard a guess that I probably got trigger happy and put it in the CI script when I noticed it wasn't in `FEATURES` - just a guess though :)

apoelstra:
I think probably when we moved the MSRV to 1.41 we noticed that schemars "worked" and didn't think about it.

Kixunil (approved):
ACK 05f543c94e6c06f07560c35fbf0f404a21fdc000

apoelstra:
lol who actually has authority to ACK PRs in this crate?

I wouldn't mind adding @Kixunil or @tcharding if they want.

apoelstra:
Added both of you with "Write" permissions.

